PR #17, opened 2015-02-02
Fix (some, few) warnings

francesco-romano:
In this patch I did the following things:
- split in the `CMakeLists.txt` the inclusion to be `SYSTEM` and not `SYSTEM`. This should help in cleaning compilers warnings.
- Cleaned up inclusions in `include/iCub/iDynTree/iDyn2KDL.h`. I did not removed redundant inclusions. Just moved them in the `cpp` and used forward declarations where possible.
- Tried some changes in unsigned vs signed variables. I tried to change the least. Hoping I did not change too much.
- Fixed some documentation warnings.


francesco-romano:
It should now build.
Tests fail on clang for some strange reason (but build should be successful)

PR #104, opened 2015-11-29
Feature/predict sensor measurement

naveenoid:
Tracking the creation of predict sensors measurement class.


src/sensors/include/iDynTree/Sensors/PredictSensorsMeasurements.hpp
```diff
@@ -0,0 +1,48 @@
...
+        bool setSensorList(const SensorsList &sensorsList);
+        bool setModel(const iDynTree::Model &model);
+        bool setTraversal(const iDynTree::Traversal &traversal);
+        bool makePrediction(iDynTree::VectorDynSize &predictedMeasurement);
+        ~PredictSensorsMeasurements();
+        
+    private :
+        
+        struct PredictSensorsMeasurementsPrivateAttributes;
+        PredictSensorsMeasurementsPrivateAttributes * pimpl;
```
traversaro: I introduced the  PIMPL in some classes to avoid having to include legacy KDL class in the headers , but in this case I guess you can also avoid it, as you prefer.

naveenoid: I tried to keep it that way because its cleaner and its unnecessary for the
user to understand that it encodes a Model, Traversal and a SensorList. On
the other hand, that is made obvious by the forward declarations..Anyways,
will code in the mapping function to the ForwardKinematics and organise my
work in a few pushes.


On Sun, Nov 29, 2015 at 6:10 PM, Silvio Traversaro <notifications@github.com

naveenoid:
@traversaro @iron76 Moving towards producing tests now...

naveenoid:
Test works for Predict SensorMeasurement (two link 1Dof arm)
Experiments : 
 1-accelerometer gravity test,
 2-angularVelocity test, 
3-AngularAccelerationTest

traversaro:
@naveenoid There seems to be an error in Travis in building (see https://travis-ci.org/robotology/idyntree/jobs/95885028): 
~~~
[ 33%] Building CXX object src/sensors/CMakeFiles/idyntree-sensors.dir/src/Sensors.cpp.o

In file included from /home/travis/build/robotology/idyntree/src/sensors/src/Sensors.cpp:30:0:

/home/travis/build/robotology/idyntree/src/sensors/include/iDynTree/Sensors/SixAxisFTSensor.h:28:40: fatal error: iDynTree/Sensors/Sensors.hpp: No such file or directory
~~~

Can you check it? 
Furthermore can you merge or rebase your branch? As you may see it is not possible to merge the pull request because there are conflicts between this branch and the master.

src/model_io/urdf/CMakeLists.txt
```diff
@@ -50,6 +50,9 @@ include_directories(${IDYNTREE_TREE_INCLUDE_DIRS})
 include_directories(SYSTEM ${urdfdom_headers_INCLUDE_DIRS}
                            ${urdfdom_INCLUDE_DIRS})
 
+if(IDYNTREE_COMPILE_TESTS)
+    add_subdirectory(tests)
+endif()
```
traversaro: If you move the `add_subdirectory(tests)` here, how can the test link to the `idyntree-modelio-urdf`that still needs to get added?

src/sensors/CMakeLists.txt
```diff
@@ -54,3 +55,10 @@ install(TARGETS ${libraryname}
         PRIVATE_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/iDynTree/Sensors/Impl)
 
 set_property(GLOBAL APPEND PROPERTY ${VARS_PREFIX}_TARGETS ${libraryname})
+
+
+# # sensors tests                            
+# if(IDYNTREE_COMPILE_TESTS)
+#     add_subdirectory(tests)
+# endif()                             
+#
```
traversaro: ?

src/sensors/src/Accelerometer.cpp
```diff
@@ -139,6 +138,20 @@ SensorType Accelerometer::getSensorType() const
     return ACCELEROMETER;
 }
 
+bool Accelerometer::getLinkSensorTransform(Transform& link_H_sensor) const
+{
+    if( this->pimpl->parent_link_index < 0)
```
traversaro: Are we sure that we want to check this in the getter ? `parent_link_index` is not used at all in this method, so if a user creates a new Accelerometer, even if I don't set the parent link index I would expect to be able to set/get the link/sensor transform.

src/sensors/src/Gyroscope.cpp
```diff
@@ -138,9 +139,12 @@ int Gyroscope::getParentIndex() const
 
 bool Gyroscope::getLinkSensorTransform(iDynTree::Transform& link_H_sensor) const
 {
+    if(this->pimpl->parent_link_index<0)
```
traversaro: See https://github.com/robotology/idyntree/pull/104/files#r47211313

src/sensors/src/PredictSensorsMeasurements.cpp
```diff
@@ -0,0 +1,185 @@
...
+{
+    
+    bool returnVal = true;
+    // incorporating gravity into the base LinAcceleration
+    iDynTree::LinkPositions linkPos(model);
+    iDynTree::LinkVelArray linkVel(model);
+    iDynTree::LinkAccArray linkAcc(model);
+    
+    
+    iDynTree::AngAcceleration nullAngAccl;
```
traversaro: Can you explicitly set the iDynTree::AngAcceleration to zero? As part of the performance improvement now vectors are not initialized to zero. See https://github.com/robotology/idyntree/issues/98#issuecomment-158999757 .

src/sensors/src/PredictSensorsMeasurements.cpp
```diff
@@ -0,0 +1,185 @@
...
+            iDynTree::Twist localVelocity;
+            Accelerometer *accelerometer = (Accelerometer*)acceleroMap[i];
+            returnVal = accelerometer->getLinkSensorTransform(link_T_sensor);
+            localVelocity = link_T_sensor*linkVel(i);
+            linkAccTerm = link_T_sensor*(linkAcc(i));
+            crossProdTerm = (localVelocity.getAngularVec3()).cross(localVelocity.getLinearVec3());
+            measuredAcc = (linkAccTerm).getLinearVec3() + crossProdTerm;  
+            predictedMeasurement.setVal(vecItr++,measuredAcc.getVal(0));
+            predictedMeasurement.setVal(vecItr++,measuredAcc.getVal(1));
+            predictedMeasurement.setVal(vecItr++,measuredAcc.getVal(2));
```
traversaro: setVal/getVal are the accessor with boundary checking (see http://wiki.icub.org/codyco/dox/html/idyntree/html/classiDynTree_1_1IVector.html#ae5e82c548aa91a66365888748f4b22b7), while the usual `operator()` is the one without boundary checking. I think in this case it make sense to use the accessor without boundary checking (even because they are much more readable).

src/sensors/src/PredictSensorsMeasurements.cpp
```diff
@@ -0,0 +1,185 @@
...
+        iDynTree::Transform link_T_sensor;
+        returnVal = gyroscope->getLinkSensorTransform(link_T_sensor);
+        measuredAngVel = (link_T_sensor*linkVel(parentLinkId)).getAngularVec3();
+        predictedMeasurement.setMeasurement(iDynTree::GYROSCOPE,idx,measuredAngVel);   
+        
+    }
+    return(returnVal);
+}
+
+bool PredictSensorsMeasurements::operator()(const Model& model,const Traversal& traversal,const iDynTree::FreeFloatingPos& robotPos,const iDynTree::FreeFloatingVel& robotVel,iDynTree::FreeFloatingAcc& robotAcc,const LinAcceleration& gravity,const iDynTree::SensorsList &sensorsList,iDynTree::VectorDynSize &predictedMeasurement)
```
traversaro: Why `operator()` ? If we want to have `PredictSensorsMeasurements` as a function (that make sense in my opinion) we should define it as a function.

src/sensors/src/PredictSensorsMeasurements.cpp
```diff
@@ -0,0 +1,185 @@
...
+{
+    bool returnVal = true;
+    // incorporating gravity into the base LinAcceleration
+    iDynTree::LinkPositions linkPos(model);
+    iDynTree::LinkVelArray linkVel(model);
+    iDynTree::LinkAccArray linkAcc(model);
```
traversaro: This initialization allocate dynamics memory at runtime, breaking real-time safeness of the code and also making it really inefficient.

src/sensors/src/PredictSensorsMeasurements.cpp
```diff
@@ -0,0 +1,185 @@
...
+        iDynTree::LinAcceleration measuredAcc, crossProdTerm;
+        iDynTree::Transform link_T_sensor;
+        iDynTree::SpatialAcc linkAccTerm;
+        iDynTree::Twist localVelocity;
+        returnVal = accelerometer->getLinkSensorTransform(link_T_sensor);
+        localVelocity = link_T_sensor*linkVel(parentLinkId);
+        linkAccTerm = link_T_sensor*(linkAcc(parentLinkId));
+        crossProdTerm = (localVelocity.getAngularVec3()).cross(localVelocity.getLinearVec3());
+        measuredAcc = (linkAccTerm).getLinearVec3() + crossProdTerm;       
+        predictedMeasurement.setMeasurement(iDynTree::ACCELEROMETER,idx,measuredAcc);
```
traversaro: I am not a taliban of avoiding code duplication when it make sense, but this is the typical code in which all kind of bugs hide, and it is: 
* Duplicated in this method. 
* Hidden in this class: if I want to "simulate" a sensor measurement from another function/class, I need to reimplement it. 
Why we don't put this code as a method in the `Accelerometer` class?

traversaro:
General comments: 
* I think it would make sense to put sensor-specific logic in sensor-specific methods. We cannot have a single interface for this "predictSensorMeasure" methods because different sensor measure different quantities, but that would definitly clean the code and make that logic available also elsewhere in the library.
* I think it would make sense to split the main logic of the class in a **function**: 
~~~
PredictSensorMeasurements(const Model & model,
const Traversal & traversal,
const SensorList & sensors,
const LinkPositions & linkPos,
const LinkVelocities & linkVel,
const LinkAccelerations & linkProperAcc, 
const LinearMotionVelocity3 & gravity,
          SensorMeasurements );
~~~

In this way, we can embed this logic in all kind of **classes** without problems. In particular I think that from the user point of view it would make sense to have a method like this to the `DynamicsComputations` class: 
~~~
DynamicsComputations::predictSensorMeasurements(SensorMeasurements & meas);
~~~
Setting the state/computing the forward kinematics would then be done using the usual `DynamicsComputations` methods, see for example the tutorial : https://github.com/robotology/idyntree/blob/master/doc/dcTutorialCpp.md . 

We can still a **class** that simplifies the use of `predictSensorMeasurements` function, but that class should then properly allocate the buffers to avoid allocating memory while "predicting measures", and should not mimic a "function" interface. 

* I think that the "serialization" of a `SensorMeasurements` to a `VectorDynSize` can be handled by an appropriate method of `SensorMeasurements`, something like `SensorMeasurements::toVector(VectorDynSize & sensVec)`, so we can avoid duplicating our function/methods.

naveenoid:
:+1

src/sensors/src/PredictSensorsMeasurements.cpp
```diff
@@ -0,0 +1,185 @@
...
+    for(idx = 0; idx<numGyro; idx++)
+    {
+        sensor = sensorsList.getSensor(iDynTree::GYROSCOPE, idx);
+        gyroMap[sensor->getParentIndex()] = sensor;
+    }
+    
+    unsigned int vecSize = (acceleroMap.size()+ gyroMap.size()) * 3;
+    unsigned int vecItr = 0;
+    predictedMeasurement.resize(vecSize);
+    for(int i =0; i<model.getNrOfDOFs();i++)
```
traversaro: If you want to iterate on links, you should use the model.getNrOfLinks() method, see http://wiki.icub.org/codyco/dox/html/idyntree/html/classiDynTree_1_1Model.html#details .

src/sensors/src/PredictSensorsMeasurements.cpp
```diff
@@ -0,0 +1,185 @@
+/*
+ * Copyright (C) 2015 Fondazione Istituto Italiano di Tecnologia
+ * Author: Naveen Kuppuswamy
+ * email:  naveen.kuppuswamyt@iit.it
```
traversaro: typo in email

src/sensors/src/PredictSensorsMeasurements.cpp
```diff
@@ -0,0 +1,185 @@
...
+    unsigned int numAccl = sensorsList.getNrOfSensors(iDynTree::ACCELEROMETER);
+    unsigned int numGyro = sensorsList.getNrOfSensors(iDynTree::GYROSCOPE);
+    int idx;
+    
+    //Iterate through each kind of accelrometer and find its parent. Compute local (classical accelration) 
+    // It is automatically proper acceleration since gravity is incorporated into the base acceleration
+
+    
+    std::map<unsigned int,Sensor*> acceleroMap;
+    std::map<unsigned int,Sensor*> gyroMap;
```
traversaro: Why instead of building a map at runtime (dynamic memory allocation) we don't simply iterate on **sensors** instead of iterating on **links** ?

naveenoid:
Few more issues from your comments to be addressed but I have rebased and moved computations to the Accelerometer and Gyroscope classes themselves

traversaro:
@naveenoid apparently you have merged against an older version of the master branch (that you probably had in the local version of the repo), and for that reason there are still conflicts. 
Before merging remember to fetch the latest version of the master branch from the `origin` repo, as described by GitHub in the instructions on how to resolve a conflict: 
mergegithub

naveenoid:
:+1:

src/sensors/CMakeLists.txt
```diff
@@ -27,11 +38,10 @@ set(libraryname idyntree-sensors)
...
 
 target_include_directories(${libraryname} PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
-                                                "$<INSTALL_INTERFACE:${CMAKE_INSTALL_PREFIX}/${${VARS_PREFIX}_INSTALL_INCLUDEDIR}>")
-target_include_directories(${libraryname} PRIVATE ${EIGEN3_INCLUDE_DIR}
-                                                  ${orocos_kdl_INCLUDE_DIRS})
+                                                "$<INSTALL_INTERFACE:${CMAKE_INSTALL_PREFIX}/${${VARS_PREFIX}_INSTALL_INCLUDEDIR}>"
+                                          PRIVATE ${EIGEN3_INCLUDE_DIR})
 
-target_link_libraries(${libraryname} idyntree-core idyntree-kdl)
+target_link_libraries(${libraryname} idyntree-core)
```
traversaro: Also idyntree-kdl needs to be added back.

src/sensors/CMakeLists.txt
```diff
@@ -44,7 +55,7 @@ install(TARGETS ${libraryname}
         PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/iDynTree/Sensors
         PRIVATE_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/iDynTree/Sensors/Impl)
 
-set_property(GLOBAL APPEND PROPERTY ${VARS_PREFIX}_TARGETS ${libraryname})
-
+#set_property(GLOBAL APPEND PROPERTY ${VARS_PREFIX}_TARGETS ${libraryname})
```
traversaro: Why this line have been commented?

traversaro:
Travis is still failing due to a failure in a test running with valgrind. 
In particular the (first) error is ( I installed valgrind with `sudo apt-get install valgrind` and enabled the CMake advanced option `IDYNTREE_RUN_VALGRIND_TESTS`, and compiled in Debug to got the debug symbols ) : 
~~~
    Start 28: memcheck_UnitTestPredictSensorsMeasurement

28: Test command: /usr/bin/valgrind "--leak-check=full" "--error-exitcode=1" "./PredictSensorsMeasurementUnitTest"
28: Test timeout computed to be: 1500
28: ==19636== Memcheck, a memory error detector
28: ==19636== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
28: ==19636== Using Valgrind-3.10.0.SVN and LibVEX; rerun with -h for copyright info
28: ==19636== Command: ./PredictSensorsMeasurementUnitTest
28: ==19636== 
28: Model /home/pegua/src/iDynTree-superbuild/libraries/iDynTree/src/tests/data/twoLinks.urdf created with :1 DoFs
28: ==19636== Conditional jump or move depends on uninitialised value(s)
28: ==19636==    at 0x567F4E7: __printf_fp (printf_fp.c:400)
28: ==19636==    by 0x567E792: vfprintf (vfprintf.c:1660)
28: ==19636==    by 0x56A2578: vsnprintf (vsnprintf.c:119)
28: ==19636==    by 0x5199E0F: ??? (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.19)
28: ==19636==    by 0x51A0263: std::ostreambuf_iterator<char, std::char_traits<char> > std::num_put<char, std::ostreambuf_iterator<char, std::char_traits<char> > >::_M_insert_float<double>(std::ostreambuf_iterator<char, std::char_traits<char> >, std::ios_base&, char, char, double) const (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.19)
28: ==19636==    by 0x51A054F: std::num_put<char, std::ostreambuf_iterator<char, std::char_traits<char> > >::do_put(std::ostreambuf_iterator<char, std::char_traits<char> >, std::ios_base&, char, double) const (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.19)
28: ==19636==    by 0x51ABAF4: std::ostream& std::ostream::_M_insert<double>(double) (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.19)
28: ==19636==    by 0x405491: iDynTree::VectorFixSize<3u>::toString() const (VectorFixSize.h:230)
28: ==19636==    by 0x403CF5: runTest(int const&, iDynTree::Model const&, iDynTree::Traversal const&, iDynTree::SensorsList const&, iDynTree::SensorsMeasurements&) (PredictSensorsMeasurementUnitTest.cpp:96)
28: ==19636==    by 0x404DE5: main (PredictSensorsMeasurementUnitTest.cpp:139)
~~~
The problem (use of uninitialized memory) seems at line 96 of file `PredictSensorsMeasurementUnitTest.cpp`, see https://github.com/robotology/idyntree/pull/104/files#diff-7647714b98c5e5be0c7ec2e846974f85R96 ) . 
It seems that makePrediction is not correctly setting the measurement, or something goes wrong in the `SensorMeasurements` class.

traversaro:
I suspect that the valgrind error is due to the check in https://github.com/robotology/idyntree/pull/104/files#diff-82a853089ea61d6ae70ac7bb71d5a487R160 in `LinAcceleration Accelerometer::predictMeasurement(const SpatialAcc& linkAcc, const iDynTree::Twist& linkTwist)`. My impression is that it would simply make sense to drop that check, as in https://github.com/robotology/idyntree/pull/104#discussion_r47211313 .

src/sensors/src/Accelerometer.cpp
naveenoid: Well the problem is without a setted 'legal' parent link index, how would one know which is the reference frame to which the link / sensor transform is being computed?

traversaro: If the code needs that information, it will call the getter for the parent_link_index, and it will get the invalid value for it.

traversaro: The method had the signature with the `bool` because it was used in the FT sensors, where the users was passing to the function the link index (one of the two that are attached to the FT sensor) with respect to which he wanted the transform. In this case the method is just returning an attribute of the class, so I think that most users don't expect all the getters of a class to stop working if an attribute as an invalid value.

src/sensors/src/PredictSensorsMeasurements.cpp
naveenoid: What is the alternative to this? I need these variables to make the ForwardKinematics call

traversaro: What we normally do (see for example the `DynamicsComputations` class) is that the Model is set with the constructor or a method intended to be used as **configuration** time. When you set the model, you can resize any dynamic sized buffer. Then the method intended to be called in the control/estimation loop uses the buffer resized at configuration time. 
On the top of that, I would also suggest to wrap all the computations for predicting sensors measurement in a function that takes in input everything (even the buffers used), so the computation can be used in several classes, for example this task-specific class or some more generic class (could be `DynamicsComputations` itself), as already discussed in https://github.com/robotology/idyntree/pull/104#issuecomment-163581506 .

traversaro:
@naveenoid any updates? Once we merge this PR we can go on and merge also https://github.com/robotology/idyntree/pull/103/files .

naveenoid:
@traversaro check src/tests/data/icub_sensorised.urdf 
It is a sensorised version of iCubGenova02. For convinience in debugging I moved all the links of the MTB (i..e the frames) which I had converted to sensor tags to the last part of the URDF.

naveenoid:
Trying one last bunch of things to understand why memcheck still fails on the predictSensorsMeasurement. I have a feling it deals with the initialisation of variables because the UnitTest on PredictMeasurement itself itsnt failing, only this memcheck.

src/model_io/urdf/tests/PredictSensorsMeasurementUnitTest.cpp
```diff
@@ -0,0 +1,146 @@
...
+                robotVel.jointVel()(0) = 0;          
+                robotAcc.jointAcc()(0) = acclTestVal;
+                break;
+    }
+    
+    predictedSensors.makePrediction(model,traversal,robotPos,robotVel,robotAcc,linkPos,linkVel,linkAcc,gravity,sensorsList,predictedMeasurement);
+    predictedMeasurement.getMeasurement(ACCELEROMETER,0,accl1);
+    predictedMeasurement.getMeasurement(ACCELEROMETER,1,accl2);
+    predictedMeasurement.getMeasurement(GYROSCOPE,0,gyro1);
+    std::cout<<"Predicted Measurement (accl1): " <<accl1.toString()<<"\n";
```
naveenoid: @traversaro for some reason this doesnt trigger a memcheck error if I use reservedToString() instead of toString()

Even though those functions are identical

src/model_io/urdf/tests/PredictSensorsMeasurementUnitTest.cpp
```diff
@@ -0,0 +1,146 @@
...
+    std::cout<<"Predicted Measurement (accl2): " <<accl2.toString()<<"\n";
+    std::cout<<"Predicted Measurement (gyro1): " <<gyro1.toString()<<"\n";
+    
+    //checking obtained results
+    switch(expID)
+    {
+        case 1 :ASSERT_EQUAL_DOUBLE(gyro1(0),0);
```
naveenoid: @traversaro all the following lines trigger memcheck errors but its due to the way vectorFixSize seems to implement the operator( )

naveenoid:
Can we merge as soon as travis is happy?

traversaro:
When Travis and AppVeyor are happy! :D 
Apparently there is still some errors in AppVeyor : 
~~~
C:\projects\idyntree\build\iDynTree.sln" (default target) (1) ->
"C:\projects\idyntree\build\ALL_BUILD.vcxproj.metaproj" (default target) (2) ->
"C:\projects\idyntree\build\src\sensors\idyntree-sensors.vcxproj.metaproj" (default target) (34) ->
"C:\projects\idyntree\build\src\sensors\idyntree-sensors.vcxproj" (default target) (35) ->
(ClCompile target) -> 
  c:\projects\idyntree\src\sensors\src\sensors.cpp(284): error C4716: 'iDynTree::SensorsMeasurements::resize' : must return a value [C:\projects\idyntree\build\src\sensors\idyntree-sensors.vcxproj]
~~~

src/sensors/src/PredictSensorsMeasurements.cpp
traversaro: @naveenoid can I go on and solve this issue?

traversaro: Actually all this method needed to be removed because it is just (untested) code duplication of the other `makePrediction` method, as discussed in the last point in https://github.com/robotology/idyntree/pull/104#issuecomment-163581506 . If you are ok with it I can work on this.

traversaro:
@naveenoid can you review the commit https://github.com/robotology/idyntree/commit/dd821dcced4cc36db57ff3ee53e0a1e772ae6429 ?

traversaro:
@naveenoid perhaps the version without the whitespace changes is easier to review : https://github.com/robotology/idyntree/commit/dd821dcced4cc36db57ff3ee53e0a1e772ae6429?w=1

naveenoid:
Sorry silvio. Im looking at the issues now. If you havent solved it yet go
ahead. Or else I will do so right now.

naveenoid:
I think we are good to merge no? I can then work on the yarpWholeBodySensors connection with the new icub URDF.

naveenoid:
Fixed the VectorDynSize implementation and added a test in the existing unitTest to verify it works. I can change documentation to update the final decisions.

traversaro:
Bindings generation fixed in https://github.com/robotology/idyntree/commit/4e3136e7c21eab6a26b882eb175a034b6fbd9129 , merging.

PR #131, opened 2016-01-25
[travis] Test regressions on issue #52

traversaro:


traversaro:
This PR is ill-posed because on travis builds iDynTree and its dependencies do not share the install include directories. We should test this on codyco-superbuild builds.

PR #152, opened 2016-04-17
[travis] Use trusty and test also examples

traversaro:
Depends on https://github.com/robotology/codyco-superbuild/pull/121 .


traversaro:
Finally green. : )

PR #281, opened 2017-04-11
Various fixes

francesco-romano:
Various fixes
- documentation
- static analyzer issues

(WIP there are other issues that can be grouped in this PR, but @traversaro you can start reviewing this commits)

### DIFF
--- a/src/core/src/MatrixDynSize.cpp
+++ b/src/core/src/MatrixDynSize.cpp
@@ -7,6 +7,7 @@
 
 #include <iDynTree/Core/MatrixDynSize.h>
 #include <iDynTree/Core/Utils.h>
+#include <algorithm>
 #include <cassert>
 #include <cstdio>
 #include <sstream>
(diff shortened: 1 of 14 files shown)


src/core/src/MatrixDynSize.cpp
```diff
@@ -76,37 +77,32 @@ MatrixDynSize& MatrixDynSize::operator=(const MatrixDynSize& other)
...
-    // Copy the new rows
+    // Copy the new rows and columns
     m_rows = other.m_rows;
     m_cols = other.m_cols;
+    unsigned requiredCapacity = m_rows * m_cols;
+
+    // if other is empty, return
+    //FIXME: This probably changed the behaviour.
+    // the capacity is never changed if the other matrix is zero
+    if (requiredCapacity == 0) return  *this;
```
traversaro: I understand this is necessary to avoid passing a null pointer to 
~~~
std::memcpy(this->m_data, other.m_data, requiredCapacity * sizeof(double));
~~~
but I think it would more readable if we just put an if check before that memcpy, to avoid having multiple return points. 
Furthermore, I do not understand the comment. Which behaviour changed?

francesco-romano: Honestly I thought the contrary. It is usually better to do "early return" as it simplify code reading.

About the previous behaviour. I think that before, if `other` is  empty, then the internal buffer is freed, while now the capacity is not changed (which is usually a desired feature)

traversaro: > Honestly I thought the contrary. It is usually better to do "early return" as it simplify code reading.

Ok, then it is a personal taste. Nevermind, leave it like that. 

> About the previous behaviour. I think that before, if other is empty, then the internal buffer is freed, while now the capacity is not changed (which is usually a desired feature)

In the previous version (see https://github.com/robotology/idyntree/pull/281/files/6e6492904cf5915d28414dfa379845dba0d6aec1#diff-c1bdcea78a4cb86a68ea5b6a1dfccccfL87) if `m_capacity` was different from zero and m_rows*m_cols was zero the if would have been skipped, so I think the behaviour is the same modulo the bug that you fixed. I think we can drop the FIXME comment.

traversaro:
For me this PR can be merged, @francesco-romano you have specific motivations behind the `[WIP]` tag?

francesco-romano:
Yes. 
The idea was to collect more fixes to avoid creating too many PR, e.g. adding the column major support to iDynTree matrices, etc, as these fixes are not urgent. 

Anyway if you prefer to merge these two we can open other PRs later

Inviato da iPhone

src/core/src/MatrixDynSize.cpp
francesco-romano: yes you are write.
I will drop that line

francesco-romano:
Let's avoid to postpone this PR.
As soon as travis is happy I will merge it.

traversaro:
Ok, merging.

francesco-romano:
I forgot 🤣

PR #299, opened 2017-05-16
Fix sensor list

francesco-romano:


francesco-romano:
@traversaro policy for `devel`? Should I merge it right now, or we can merge master into devel later?

PR #410, opened 2018-01-28
Adding Schmitt Trigger based contact state detection for Contact Swit…

prashanthr05:
preliminary version of contact switching

To determine the primary foot in contact with the ground, based on a alternate contact switching pattern, by performing a threshold over a contact make and contact break normal force magnitudes using a Schmitt trigger.

* Schmitt Trigger class takes 4 parameters (low threshold, high threshold, stable off time, stable on time) to determine the current binary state of the device, based on current time and an input value. 

* Contact state machine class contains an object of the Schmitt trigger specific to contact state detection and determines the contact transition modes.

* Foot contact classifier class considers contact state machines for each foot (left and right, in case of a humanoid)  and determines the contact state of each foot to set the primary foot in contact. This information can further be used by the legged odometry(LO) device to change fixed frame of reference, since the LO module assumes that atleast one foot is in contact at any instant of time. In its current implementation, the setting for primary foot is done only in the context of an alternate contact switching pattern.

### DIFF
--- a/src/estimation/include/iDynTree/Estimation/ContactStateMachine.h
+++ b/src/estimation/include/iDynTree/Estimation/ContactStateMachine.h
@@ -1,0 +1,136 @@
+/*
+ * Copyright (C) 2018 Fondazione Istituto Italiano di Tecnologia
+ * Authors: Silvio Traversaro, Prashanth Ramadoss
+ * email: silvio.traversaro@iit.it, prashanth.ramadoss@iit.it
+ *
+ * Permission is granted to copy, distribute, and/or modify this program
+ * under the terms of the GNU General Public License, version 2 or any
+ * later version published by the Free Software Foundation.
+ *
+ * A copy of the license can be found at
+ * http://www.robotcub.org/icub/license/gpl.txt
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details
+ */
+
+#ifndef IDYNTREE_CONTACTSTATEMACHINE_H
+#define IDYNTREE_CONTACTSTATEMACHINE_H
+#include <iostream>
+#include <memory>
+#include "SchmittTrigger.h"
+
+namespace iDynTree
+{
+    /**
+     * struct to hold schmitt trigger device parameters
+     */
+    struct SchmittParams
+    {
+        double stableTimeContactMake;
+        double stableTimeContactBreak;
+        double contactMakeForceThreshold;
+        double contactBreakForceThreshold;
+    };
+    
+   /**
+    * Contact State Machine class for binary contact state detection
+    * Contains a Schmitt Trigger device for updating the contact states
+    * using the contact normal force acting on the contact link and
+    * Determines contact transitions using simple binary switching logic
+    * 
+    * Can be used to determine stable contacts, contact breaking and contact making.
+    * The parameters to the Schmitt Trigger are passed as a struct and is the Schmitt 
+    * Trigger is instantiated in the class constructor. 
+    * 
+    * NOTE: There are no default parameters to the Schmitt Trigger. These parameters are set through 
+    * the constructor during instantiation.
+    * 
+    * This class does not exactly abstract the Schmitt Trigger class. Schmitt Trigger methods are still accessible through 
+    * the m_contactSchmitt object. This class uses a generic Schmitt Trigger object and augments its functionality
+    * specific to physical contacts based scenarios.
+    */
+    class ContactStateMachine
+    {
+    public:
+        /**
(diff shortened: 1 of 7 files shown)


prashanthr05:
cc @traversaro

src/estimation/include/iDynTree/Estimation/ContactStateMachine.h
```diff
@@ -0,0 +1,139 @@
...
+ * Copyright (C) 2018 Fondazione Istituto Italiano di Tecnologia
+ * Authors: Silvio Traversaro, Prashanth Ramadoss
+ * email: silvio.traversaro@iit.it, prashanth.ramadoss@iit.it
+ *
+ * Permission is granted to copy, distribute, and/or modify this program
+ * under the terms of the GNU General Public License, version 2 or any
+ * later version published by the Free Software Foundation.
+ *
+ * A copy of the license can be found at
+ * http://www.robotcub.org/icub/license/gpl.txt
```
traversaro: I just noticed that this is a dead URL. Anyhow, I saw that you copied the header from other files, so it probably make more sense to change all the headers once.

src/estimation/include/iDynTree/Estimation/ContactStateMachine.h
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details
+ */
+
+#ifndef CONTACTSTATEMACHINE_H
```
traversaro: Can you add `IDYNTREE_` to the guard macro?

src/estimation/include/iDynTree/Estimation/ContactStateMachine.h
prashanthr05: I did not check the contents of the url! I just copy-pasted, yes ! What do you suggest to do about this ?

src/estimation/include/iDynTree/Estimation/ContactStateMachine.h
```diff
@@ -0,0 +1,139 @@
...
+#define CONTACTSTATEMACHINE_H
+#include <iostream>
+#include <memory>
+#include "SchmittTrigger.h"
+
+namespace iDynTree
+{
+namespace ContactSwitching
```
traversaro: For compatibility with how these classes are exposed in bindings, we try to use a single namespace in iDyntTree, I think it is ok to just leave this classes in the `iDynTree` namespace.

src/estimation/include/iDynTree/Estimation/ContactStateMachine.h
traversaro: Note: this applies also to the other classes.

src/estimation/include/iDynTree/Estimation/ContactStateMachine.h
traversaro: Note: this applies also to the other classes.

prashanthr05: Fixed it.

src/estimation/include/iDynTree/Estimation/ContactStateMachine.h
prashanthr05: Fixed it.

src/estimation/include/iDynTree/Estimation/ContactStateMachine.h
traversaro: Leave it as that, we will cleanup the headers automatically in the future.

src/estimation/include/iDynTree/Estimation/FootContactClassifier.h
```diff
@@ -0,0 +1,150 @@
...
+    * while one of the two foot breaks contact with the ground surface, since LO assumes that atleast one
+    * foot is in contact with ground at any instant of time. However, it is also used to assign 
+    * the primary foot in case of double stance.
+    * 
+    * In the current version of this class, switching logic is implemented only for 
+    * ALTERNATE_CONTACT patterns vastly considered during walking tasks. This class is aimed
+    * to be extended towards other switching patterns like LATEST_ACTIVE_CONTACT, DEFAULT_CONTACT
+    * for implementation along side different task based controllers. 
+    */    
+    class FootContactClassifier
```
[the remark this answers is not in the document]
prashanthr05: Makes sense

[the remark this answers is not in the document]
prashanthr05: Changed it. Should I rebase and squash it into a single commit ?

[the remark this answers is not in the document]
traversaro: It would be ideal, thanks!

traversaro:
Thanks, merging.

PR #450, opened 2018-06-10
Completed angular acceleration predicted measurements implementation and added sensor bindings

nunoguedelha:
### Completed angular acceleration predicted measurements implementation
- added missing method in `predictSensorsMeasurementsFromRawBuffers`
- return pass/fail status from `setMeasurement()`
- deprecated method `predictMeasurement()` interface which required
  both link spatial acceleration and velocity. New method interface
  only requires link acceleration.

### Added bindings for sensor ThreeAxisAngularAccelerometer
- added bindings for the new sensor class
- cleaned-up extra files (iDynTree specific SwigRef.m ...)  shading generated ones
- regenerated bindings using new robotology SWIG (`matlab` branch) commit : 
   7e61db6b5dc841027439081aed3bf8c36c9e1920


nunoguedelha:
@traversaro , feel free to suggest a second reviewer...

PR #504, opened 2018-11-02
Fix getRPY documentation in Rotation.h

prajval10:
Changing limits of RPY documention:
Get `r \in [-π, π] , p \in [-π/2, π/2], y \in [-π, π]` to 
Get `r \in [-π, π] , p \in (-π/2, π/2), y \in [-π, π]`

### DIFF
--- a/src/core/include/iDynTree/Core/Rotation.h
+++ b/src/core/include/iDynTree/Core/Rotation.h
@@ -164,7 +164,7 @@ namespace iDynTree
          /**
          * Get a roll, pitch and yaw corresponding to this rotation.
          *
-         * Get \f$ r \in [-π, π] , p \in [-π/2, π/2], y \in [-π, π]\f$
+         * Get \f$ r \in [-π, π] , p \in (-π/2, π/2), y \in [-π, π]\f$
          * such that
          * *this == RotZ(y)*RotY(p)*RotX(r)
          *
--- a/src/core/include/iDynTree/Core/Rotation.h
+++ b/src/core/include/iDynTree/Core/Rotation.h
@@ -164,7 +164,7 @@ namespace iDynTree
          /**
          * Get a roll, pitch and yaw corresponding to this rotation.
          *
-         * Get \f$ (r,p,y) \in (-π, π] \times (-π/2, π/2) \times (-π, π] \bigcap  \{0\} \times \{-π/2,π/2\} \times (-π,π)\f$
+         * Get \f$ (r,p,y) \in ( (-π, π] \times (-π/2, π/2) \times (-π, π] ) \cup ( \{0\} \times \{-π/2\} \times (-π,π) ) \cup ( \{0\} \times \{π/2\} \times (-π,π) )\f$
          * such that
          * *this == RotZ(y)*RotY(p)*RotX(r)
          *
--- a/src/core/include/iDynTree/Core/Rotation.h
+++ b/src/core/include/iDynTree/Core/Rotation.h
@@ -164,7 +164,7 @@ namespace iDynTree
          /**
          * Get a roll, pitch and yaw corresponding to this rotation.
          *
-         * Get \f$ r \in [-π, π] , p \in (-π/2, π/2), y \in [-π, π]\f$
+         * Get \f$ (r,p,y) \in (-π, π] \times (-π/2, π/2) \times (-π, π] \bigcap  \{0\} \times \{-π/2,π/2\} \times (-π,π)\f$
          * such that
          * *this == RotZ(y)*RotY(p)*RotX(r)
          *
--- a/src/core/include/iDynTree/Core/Rotation.h
+++ b/src/core/include/iDynTree/Core/Rotation.h
@@ -164,7 +164,7 @@ namespace iDynTree
          /**
          * Get a roll, pitch and yaw corresponding to this rotation.
          *
-         * Get \f$ (r,p,y) \in ( (-π, π] \times (-π/2, π/2) \times (-π, π] ) \cup ( \{0\} \times \{-π/2\} \times (-π,π) ) \cup ( \{0\} \times \{π/2\} \times (-π,π) )\f$
+         * Get \f$ (r,p,y) \in ( (-π, π] \times (-π/2, π/2) \times (-π, π] ) \cup ( \{0\} \times \{-π/2\} \times (-π,π] ) \cup ( \{0\} \times \{π/2\} \times [-π,π) )\f$
          * such that
          * *this == RotZ(y)*RotY(p)*RotX(r)
          *


traversaro:
Are you sure that it is `r \in [-π, π] , p \in (-π/2, π/2), y \in [-π, π]` ? 
I would have expected that for each range had one side that is [open](https://en.wikipedia.org/wiki/Open_set#Open_and_closed_are_not_mutually_exclusive) and one side that is closed.

prashanthr05:
Open at one side makes sense. In degrees, 0 -> 179 -> -180 ->0.

prajval10:
In my tests, `pitch` doesn't seem to work at both `pi/2` and `-pi/2`. So I think it must be open on both sides, also suggested by [this (page 5 "either case")](http://www.gregslabaugh.net/publications/euler.pdf). However [this](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=1301) suggests open on one side and closed on the other. Not conclusive though.

prashanthr05:
IMO pitch should be open on both sides, because its a singularity.

traversaro:
I think it boils down to the ranges of the functions used to compute r,p and y: 
https://github.com/robotology/idyntree/blob/c8bf721b771fa4b1e7c3a940632e121060719a19/src/core/src/Rotation.cpp#L311 .

Looking at the code now, I wonder about the logic: being a product of `sin` and `cos`, the elements of a rotation matrix can never been greater than `1.0`, so the two ifs actually implicitly check (with a tolerance just the upper side) if R(2,0) is equal to 1.0 or -1.0, and assign `p` to pi/2 and -pi/2 . Probably in your tests @prajval10  `R(2,0)` is never exactly 1.0?

traversaro:
Regarding the other two (r and y) `std::atan2` apparently is defined as returning values in  (-π , +π] (see https://en.cppreference.com/w/cpp/numeric/math/atan2), so also r and y should have similar ranges.

prashanthr05:
> so the two ifs actually implicitly check (with a tolerance just the upper side) if R(2,0) is equal to 1.0 or -1.0, and assign p to pi/2 and -pi/2

So in this case, it's a closed set including pi/2  and -pi/2 ?

lrapetti:
Just as a clarification, 
I have seen that from a given `RPY(roll, pitch, yaw)` the rotation is computed [as follow](https://github.com/robotology/idyntree/blob/c8bf721b771fa4b1e7c3a940632e121060719a19/src/core/src/RotationRaw.cpp#L184):
```
compose(RotZ(yaw), compose(RotY(pitch), RotX(roll)));
```
which I got that means we are rotating around the **fixed frame** XYZ axis with the followind order: X -> Y -> Z (which is equivalant to rotate around the **moving frame** X'Y'Z' with the opposite order Z' -> Y' -> X'). 
Am I correct? Because usally what I see is that they are used the other way around.

prashanthr05:
@lrapetti  yes! Matrix multiplications must be read from right to left. In that sense, we first rotate about X, then Y and then Z.

lrapetti:
> @lrapetti yes! Matrix multiplications must be read from right to left. In that sense, we first rotate about X, then Y and then Z.

Yes yes, I meant that I was expecting the rotations to be in the opposite order accordingly to definitions I have found.
I mostly saw it defined as follow (but that's just about chosen convention)
screen shot 2018-12-11 at 17 59 41

In my case then I most likely may require to get the Euler angles defined in the opposite order. I could do something like `getYPR()`, but I fear that it may increase the confusion here. What would you suggest?

prajval10:
> This is exactly the convention used by KUKA if I am not wrong @prajval10 

Yes this is correct. To sum up, as discussed in [this issue](https://github.com/iit-danieli-joint-lab/idjl-software/issues/89), IDynTree and most other conventions:
`RPY(roll, pitch, yaw) = (RotZ(yaw) * RotY(pitch) * RotX(roll))`
Kuka's convention (as described in manual):
`RPY(roll, pitch, yaw) = (RotZ(roll) * RotY(pitch) * RotX(yaw))`

lrapetti:
> To be honest, I am unable to process the phrase "fixed frame/moving frame" if they are not precisely defined mathematically, at least using a reference. I typically just think of all this parametrization in term of how they are mapped to the SO(3) rotation matrix space, so in this case the definition is given by the equation of the parametrization itself, and by the subset of the R^3 assumed to ensure that the R^3 <---> SO(3) map is bijective. I don't think that trying to add a concept of "rotationing around the axis" helps in understanding more the convention used.

I was referring to the definition of extrinsic/intrinsic rotation definition (also refered as fixed/mobile).

> This is exactly the convention used by KUKA if I am not wrong @prajval10 . As long as everything is properly defined and documented in mathematical terms, I am not opposed to add more methods, as long as they are useful. We can also look if it make sense to add a flexible conversion method such as the one present in the Eigen EulerAngles unsupported module: https://eigen.tuxfamily.org/dox/unsupported/classEigen_1_1EulerSystem.html . However, if such a solution is used, I would really prefer to have the precise definition of the resulting R^3 <---> SO(3) map in the documentation, instead of the abract terms such as " tait bryan" or "proper/classic Euler angles".

Yes, a proper documentation here is important given the different conventions for defining roll pitch yaw and euler angles.

DanielePucci:
Can we merge this? 

CC
@prajval10 @traversaro

traversaro:
The [image](https://en.wikipedia.org/wiki/Image_(mathematics)) of the GetRPY function is actually more complicated that the one proposed in this PR by @prajval10 . We can merge the PR if it is updated to reflect the actual image of the `GetRPY` function.

src/core/include/iDynTree/Core/Rotation.h
```diff
@@ -164,7 +164,7 @@ namespace iDynTree
          /**
          * Get a roll, pitch and yaw corresponding to this rotation.
          *
-         * Get \f$ r \in [-π, π] , p \in [-π/2, π/2], y \in [-π, π]\f$
+         * Get \f$ (r,p,y) \in (-π, π] \times (-π/2, π/2) \times (-π, π] \bigcap  \{0\} \times \{-π/2,π/2\} \times (-π,π)\f$
```
traversaro: Modifications:
* Are you use that you want to use the "intersection" of sets? I think we should have a union of sets there, so `\bigcap` --> `\bigcup` 
* As all the symbols are regular size, I think it is more readable if also the set union is regular size, so `\bigcup` --> `\cup`. 
* Add parenthesis to be explicit on the precedence between the cartesian product operator and the set union operator (I guess most users are not aware about the consensus about precedence between these operators)
* In the configuration set corresponding to singularity, I guess the yaw can also take the exact value `\pi`, as it is computed using the `atan2` function, right? So I added `]` in the last `(-\pi, \pi)`.
See the following suggestion:
```suggestion
         * Get \f$ (r,p,y) \in ((-π, π] \times (-π/2, π/2) \times (-π, π]) \cup ( \{0\} \times \{-π/2,π/2\} \times (-π,π] )\f$
```

prajval10: Yes sorry, it is the union. Also according to the source code:
```
if (R(2, 0) < 1.0) {
        if (R(2, 0) > -1.0) {
            r = atan2(R(2, 1), R(2, 2));
            p = asin(-R(2, 0));
            y = atan2(R(1, 0), R(0, 0));
        } else {
            // Not a unique solution
            // Singularity
            r = 0.0;
            p = IDJL_M_PI / 2.0;
            y = -atan2(-R(1, 2), R(1, 1));
        }
    } else {
        // Not a unique solution
        // Singularity
        r = 0.0;
        p = -IDJL_M_PI / 2.0;
        y = atan2(-R(1, 2), R(1, 1));
    }
```
We should have two separate cases handling the singularity.

traversaro: > We should have two separate cases handling the singularity.

Indeed. That may be nice, as the `\{-π/2,π/2\}` part was not super understandable in any case.

traversaro:
Thanks @prajval10 .

PR #548, opened 2019-08-01
Feature: yarprobotstatepublisher reduced model tf optional parameter

Yeshasvitvs:
Currently, `yarprobotstatepublisher` publishes the transforms (TFs) of [all the frames](https://github.com/robotology/idyntree/blob/devel/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp#L242) present in a model passed as argument. This frames include also the frames that are associated with sensors e.g. [skin sensor frame](https://github.com/robotology/icub-models/blob/master/iCub/robots/iCubGenova04/model.urdf#L186). These frames without any geometrical properties are not necessary for rviz visualization and we can safely ignore to send the TF associated with these frames from `yarprobotstatepublisher`. This will help to speed up rviz visualization and solves [problems like drop in frame rate](https://github.com/robotology/human-dynamics-estimation/issues/134) due to huge number of TFs present in the transform server. In case of iCub, the total number of frames (including the sensor frames) are **262** while the total number of links that have geometrical properties for rviz visualization are only **39**. 

This PR adds an addition optional boolean parameter `--reduced-model` to `yarprobotstatepublisher`. The default value is **false** and `yarprobotstatepublisher` streams TFs of all the frames present in the urdf model. If set to **true**, only the TFs of links are streamed.

@traversaro @diegoferigo @lrapetti @DanielePucci

### DIFF
--- a/src/tools/yarprobotstatepublisher/src/main.cpp
+++ b/src/tools/yarprobotstatepublisher/src/main.cpp
@@ -38,8 +38,8 @@ int main(int argc, char *argv[])
         cout<<"\t--name-prefix            <name-prefix> : prefix of the yarprobotstatepublisher ports (default no prefix)"<<endl;
         cout<<"\t--tf-prefix              <tf-prefix>   : prefix of the published TFs (default no prefix)"<<endl;
         cout<<"\t--model                  <file-name>   : file name of the model to load at startup"<<endl;
-        cout<<"\t--reduced-model          <boolean-flag>: set true to stream only the link TFs\n"
-              "                                           \t The default value is false and TFs of all the frames in the model are streamed"<<endl;
+        cout<<"\t--reduced-model          <boolean-flag>: use the option to stream only the link TFs\n"
+              "                                           \t By default TFs of all the frames in the model are streamed"<<endl;
         cout<<"\t--base-frame             <frame-name>  : specify the base frame of the published tf tree"<<endl;
         cout<<"\t--jointstates-topic      <topic-name>  : source ROS topic that streams the joint state (default /joint_states)"<<endl;
         return EXIT_SUCCESS;
--- a/src/tools/yarprobotstatepublisher/src/main.cpp
+++ b/src/tools/yarprobotstatepublisher/src/main.cpp
@@ -35,11 +35,13 @@ int main(int argc, char *argv[])
     if (rf.check("help"))
     {
         cout<<"Options"<<endl;
-        cout<<"\t--name-prefix            <name-prefix>: prefix of the yarprobotstatepublisher ports (default no prefix)"<<endl;
-        cout<<"\t--tf-prefix              <tf-prefix>: prefix of the published TFs (default no prefix)"<<endl;
-        cout<<"\t--model                  <file-name>: file name of the model to load at startup"<<endl;
-        cout<<"\t--base-frame             <frame-name>: specify the base frame of the published tf tree"<<endl;
-        cout<<"\t--jointstates-topic      <topic-name>: source ROS topic that streams the joint state (default /joint_states)"<<endl;
+        cout<<"\t--name-prefix            <name-prefix> : prefix of the yarprobotstatepublisher ports (default no prefix)"<<endl;
+        cout<<"\t--tf-prefix              <tf-prefix>   : prefix of the published TFs (default no prefix)"<<endl;
+        cout<<"\t--model                  <file-name>   : file name of the model to load at startup"<<endl;
+        cout<<"\t--reduced-model          <boolean-flag>: set true to stream only the link TFs\n"
+              "                                           \t The default value is false and TFs of all the frames in the model are streamed"<<endl;
+        cout<<"\t--base-frame             <frame-name>  : specify the base frame of the published tf tree"<<endl;
+        cout<<"\t--jointstates-topic      <topic-name>  : source ROS topic that streams the joint state (default /joint_states)"<<endl;
         return EXIT_SUCCESS;
     }
 
--- a/src/tools/yarprobotstatepublisher/src/main.cpp
+++ b/src/tools/yarprobotstatepublisher/src/main.cpp
@@ -38,7 +38,7 @@ int main(int argc, char *argv[])
         cout<<"\t--name-prefix            <name-prefix> : prefix of the yarprobotstatepublisher ports (default no prefix)"<<endl;
         cout<<"\t--tf-prefix              <tf-prefix>   : prefix of the published TFs (default no prefix)"<<endl;
         cout<<"\t--model                  <file-name>   : file name of the model to load at startup"<<endl;
-        cout<<"\t--reduced-model          <boolean-flag>: use the option to stream only the link TFs\n"
+        cout<<"\t--reduced-model                        : use the option to stream only the link TFs\n"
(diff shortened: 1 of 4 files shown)


src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
```diff
@@ -150,6 +150,11 @@ bool YARPRobotStatePublisherModule::configure(ResourceFinder &rf)
         return false;
     }
 
+    // Set reduced model option
+    // The default value is false and considers streaming TFs of all the frames in the model
+    // If set to true, only the TFs of the links are streamed to transform server
+    this->reducedModelOption=rf.check("reduced-model",Value(0)).asBool();
```
traversaro:
```suggestion
    this->reducedModelOption=rf.check("reduced-model", Value(0)).asBool();
```

traversaro: By the way, I think it is more usable to have boolean flags just as if `--reduced-model` is present enable the option, otherwise disable it.

src/tools/yarprobotstatepublisher/src/main.cpp
```diff
@@ -35,11 +35,13 @@ int main(int argc, char *argv[])
...
         cout<<"Options"<<endl;
-        cout<<"\t--name-prefix            <name-prefix>: prefix of the yarprobotstatepublisher ports (default no prefix)"<<endl;
-        cout<<"\t--tf-prefix              <tf-prefix>: prefix of the published TFs (default no prefix)"<<endl;
-        cout<<"\t--model                  <file-name>: file name of the model to load at startup"<<endl;
-        cout<<"\t--base-frame             <frame-name>: specify the base frame of the published tf tree"<<endl;
-        cout<<"\t--jointstates-topic      <topic-name>: source ROS topic that streams the joint state (default /joint_states)"<<endl;
+        cout<<"\t--name-prefix            <name-prefix> : prefix of the yarprobotstatepublisher ports (default no prefix)"<<endl;
+        cout<<"\t--tf-prefix              <tf-prefix>   : prefix of the published TFs (default no prefix)"<<endl;
+        cout<<"\t--model                  <file-name>   : file name of the model to load at startup"<<endl;
+        cout<<"\t--reduced-model          <boolean-flag>: set true to stream only the link TFs\n"
```
traversaro: Just to understand, you have any idea why all the lines changed? Furthermore, for consistency is nice if you terminate all lines with `endl`, not with `\n` just in one case.

src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
Yeshasvitvs: @traversaro I am not following the comment correctly

src/tools/yarprobotstatepublisher/src/main.cpp
Yeshasvitvs: > Just to understand, you have any idea why all the lines changed?

Just reordered to be aligned at `:`

> Furthermore, for consistency is nice if you terminate all lines with `endl`, not with `\n` just in one case.

I use `\n` to be consistent with the `cout` usage only for the parameters but not the debug info of the parameter, If you want I can change it. The output with the current format looks 

Selection_001

src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
traversaro: Optional boolean arguments are tipically passed in a form of a flag that can be either be present or not, i.e. `--option`, not `--option true`. See for example the `ls` command in *nix systems: 
~~~
straversaro@iiticublap103:~/src/robotology-superbuild/robotology/robots-configuration/testSkin$ ls --help
Usage: ls [OPTION]... [FILE]...
List information about the FILEs (the current directory by default).
Sort entries alphabetically if none of -cftuvSUX nor --sort is specified.

Mandatory arguments to long options are mandatory for short options too.
  -a, --all                  do not ignore entries starting with .
  -A, --almost-all           do not list implied . and ..
      --author               with -l, print the author of each file
  -b, --escape               print C-style escapes for nongraphic characters
      --block-size=SIZE      scale sizes by SIZE before printing them; e.g.,
                               '--block-size=M' prints sizes in units of
                               1,048,576 bytes; see SIZE format below
  -B, --ignore-backups       do not list implied entries ending with ~
  -c                         with -lt: sort by, and show, ctime (time of last
                               modification of file status information);
                               with -l: show ctime and sort by name;
                               otherwise: sort by ctime, newest first
~~~
The options are `--all`, `--escape`, not `--all true`, `--escape`, etc etc.

Yeshasvitvs: I see what you mean, if I understand well this boolean option should not be given as `--reduced-model true` but be simply used as the argument `--reduced-model` and is passed this option should be used. :+1:

traversaro: Exactly!

Yeshasvitvs: Updated https://github.com/robotology/idyntree/pull/548/commits/9bc68bba7601c66be48c7db0bb6381805b897619

src/tools/yarprobotstatepublisher/src/main.cpp
```diff
@@ -35,11 +35,13 @@ int main(int argc, char *argv[])
...
         cout<<"Options"<<endl;
-        cout<<"\t--name-prefix            <name-prefix>: prefix of the yarprobotstatepublisher ports (default no prefix)"<<endl;
-        cout<<"\t--tf-prefix              <tf-prefix>: prefix of the published TFs (default no prefix)"<<endl;
-        cout<<"\t--model                  <file-name>: file name of the model to load at startup"<<endl;
-        cout<<"\t--base-frame             <frame-name>: specify the base frame of the published tf tree"<<endl;
-        cout<<"\t--jointstates-topic      <topic-name>: source ROS topic that streams the joint state (default /joint_states)"<<endl;
+        cout<<"\t--name-prefix            <name-prefix> : prefix of the yarprobotstatepublisher ports (default no prefix)"<<endl;
+        cout<<"\t--tf-prefix              <tf-prefix>   : prefix of the published TFs (default no prefix)"<<endl;
+        cout<<"\t--model                  <file-name>   : file name of the model to load at startup"<<endl;
+        cout<<"\t--reduced-model          <boolean-flag>: use the option to stream only the link TFs\n"
```
traversaro:
```suggestion
        cout<<"\t--reduced-model                        : use the option to stream only the link TFs\n"
```

traversaro:
Thanks, almost done. Can you add a line in the changelog at https://github.com/robotology/idyntree/blob/devel/doc/releases/v0_12.md on this? Thanks!

Yeshasvitvs:
> Thanks, almost done. Can you add a line in the changelog at https://github.com/robotology/idyntree/blob/devel/doc/releases/v0_12.md on this? Thanks!

Done https://github.com/robotology/idyntree/pull/548/commits/942eac43d22ca0b685032e7da0f50833908c0f40

traversaro:
Thanks!

PR #551, opened 2019-08-05
Feature: yarprobotstatepublisher update joints size check

Yeshasvitvs:
This PR addressed the segmentation fault that happens when trying to get the robot joint index based on the joint name present in the `/joint_states` topic.
https://github.com/robotology/idyntree/blob/2cbb75f67a8a9b93428d93d941c8d1b65ba6a415/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp#L236

On the real robot (`iCubGenova04`), the `/joint_states` topic consists of **50** joints, including the joint values of the fingers, where as the `model.urdf` of `iCubGenova04` doesn;t have the hands. So, when trying to retrieve a finger joint based from the mode, segmentation fault occurs.

Also, this PR removes the joint size check https://github.com/robotology/idyntree/blob/devel/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp#L220

@traversaro

### DIFF
--- a/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
+++ b/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
@@ -128,6 +128,9 @@ bool YARPRobotStatePublisherModule::configure(ResourceFinder &rf)
     // Resize the joint pos buffer
     m_jointPos.resize(m_kinDynComp.model().getNrOfPosCoords());
 
+    // Initilize the joint pos buffer to Zero
+    m_jointPos.zero();
+
     // Get the base frame information
     if (rf.check("base-frame"))
     {
--- a/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
+++ b/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
@@ -217,14 +217,6 @@ void YARPRobotStatePublisherModule::onRead(yarp::rosmsg::sensor_msgs::JointState
         return;
     }
 
-    // Check if joint states contain at least as many joints of the model
-    if (v.name.size() < m_jointPos.size())
-    {
-        yError() << "Size mismatch. Model has " << m_jointPos.size()
-                 << " joints, while the received JointState message has " << v.name.size() << " joints.";
-        return;
-    }
-
     // TODO: this part can be drastically speed up.
     //      Possible improvements:
     //        * Add a map string --> indeces
--- a/doc/releases/v0_12.md
+++ b/doc/releases/v0_12.md
@@ -59,3 +59,6 @@ KinDynComputations finally reached feature parity with respect to DynamicsComput
 * Add `tf-prefix` and `jointstates-topic` options for the tf prefixes and ROS topic.
 * `robot` option is deprecated, and replaced by `name-prefix`.
 * Add `reduced-model` optional parameter to stream only the link transformations to transform server. By default, tranformations from all the frames are streamed to the transform server.
+* Remove joint size check between model joints and joints in ROS topic given through `jointstates-topic` parameter.
+* Initialize the model joint positions values to zero and update the joint positions values in run time if the values are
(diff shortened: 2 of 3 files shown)


src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
```diff
@@ -233,9 +225,16 @@ void YARPRobotStatePublisherModule::onRead(yarp::rosmsg::sensor_msgs::JointState
...
-        if (!(model.getJoint(jntIndex)->getNrOfDOFs()) || jntIndex == iDynTree::JOINT_INVALID_INDEX)
+        for (size_t j=0; j < model.getNrOfDOFs(); j++)
+        {
+            if (v.name[i] == model.getJointName(j))
+            {
+                jntIndex = model.getJointIndex(v.name[i]);
+                break;
+            }
+        }
+
```
traversaro: Can't we just leave the `jntIndex = model.getJointIndex(v.name[i]);` line, and just check if `jntIndex` is `iDynTree::JOINT_INVALID_INDEX` ?
`

Yeshasvitvs: Yes, force pushed a new commit that removes only the joint size check and doesn't modify joint index handling.

Yeshasvitvs:
> To which values are joint set if they are not found in the topic ? This should be documented, otherwise it is not clear how the tool behaves in that case.

@traversaro the current logic only sets the positions of the model joints that are present in the rostopic https://github.com/robotology/idyntree/blob/2cbb75f67a8a9b93428d93d941c8d1b65ba6a415/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp#L241

So, we are not explicitly setting the position values of the model joints that are not present in the rostopic. I see that the joint positions variable is resized https://github.com/robotology/idyntree/blob/2cbb75f67a8a9b93428d93d941c8d1b65ba6a415/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp#L129

I am not certain if the values of set to  **zero** while [resizing](http://wiki.icub.org/codyco/dox/html/idyntree/html/classiDynTree_1_1VectorDynSize.html#ab10cf86d773fa67a1522c3b229c4ffc0). If not the case we can initilize the joint positions to [zero](http://wiki.icub.org/codyco/dox/html/idyntree/html/classiDynTree_1_1VectorDynSize.html#ad264c9a85bf7adbd308627b0e719d65f).

traversaro:
> I am not certain if the values of set to zero while resizing. If not the case we can initilize the joint positions to zero.

To protect ourself again future modifications of the `resize` method, I would explicitly zero the vector calling `.zero()`, and then I would document this in the command documentation, thanks!

Yeshasvitvs:
> > I am not certain if the values of set to zero while resizing. If not the case we can initilize the joint positions to zero.
> 
> To protect ourself again future modifications of the `resize` method, I would explicitly zero the vector calling `.zero()`, and then I would document this in the command documentation, thanks!

Done in https://github.com/robotology/idyntree/pull/551/commits/7d4326275b58d3cb0a700c2cb95f1984d6fd5a66

The updated commandline documentation looks like

Screenshot from 2019-08-19 15-46-56

traversaro:
Great, thanks. Can you add a line to the changelog in https://github.com/robotology/idyntree/blob/devel/doc/releases/v0_12.md ? Thanks!

Yeshasvitvs:
> Great, thanks. Can you add a line to the changelog in https://github.com/robotology/idyntree/blob/devel/doc/releases/v0_12.md ? Thanks!

Done https://github.com/robotology/idyntree/pull/551/commits/0cc039f2184a6584ee7cc4e412fee5bcc2a8906b

Yeshasvitvs:
@traversaro Ping

traversaro:
Thanks @Yeshasvitvs . Ignoring failures in Travis that are due to https://github.com/robotology/idyntree/issues/558 .

PR #646, opened 2020-01-27
Add begin/end member functions for iDynTree vectors

GiulioRomualdi:
This PR:
- Implement `begin()/end()` and `cbegin()/cend()` methods for `VectorFixSize`
- Implement `begin()/end()` and `cbegin()/cend()` methods for `VectorDynSize`
- Update the `CHANGELOG.md`

The implementation of these members will allow the user to use [range-based for loop](https://en.cppreference.com/w/cpp/language/range-for).

Related issue #645

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Added `bindings` for `InverseKinematics` (https://github.com/robotology/idyntree/pull/633)
+- Implement `cbegin()` / `cend()` and `begin()` / `end()` methods for `VectorDynSize` and `VectorFixSize` (https://github.com/robotology/idyntree/pull/646)
 
-### Fixed 
+### Fixed
 - Remove spurious inclusion of Eigen headers in ExtendedKalmanFilter.h public header, that could create probles when using that header in a downstream project that does not use Eigen (https://github.com/robotology/idyntree/pull/639).
 - Added find_dependency(OsqpEigen) and find_dependency(LibXml2) when iDynTree is compiled as a static library, fixing the use of iDynTree on Windows (https://github.com/robotology/idyntree/pull/642).
 
--- a/src/core/include/iDynTree/Core/VectorDynSize.h
+++ b/src/core/include/iDynTree/Core/VectorDynSize.h
@@ -127,6 +127,48 @@ namespace iDynTree
 
         bool setVal(const unsigned int index, const double new_el);
 
+        /**
+         * Returns a const iterator to the beginning of the vector
+         * @note At the moment iterator is implemented as a pointer, it may change in the future.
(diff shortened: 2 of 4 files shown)


traversaro:
Thanks  for the contribution @GiulioRomualdi ! Technically this is a (small) new feature, so I would prefer to merge it in `devel` and eventually release it as iDynTree 1.1 , instead of merging it in master and release it as part of iDynTree 1.0.2, to comply with SemVer (see https://semver.org/). 

However, as it is  the "saturday  that is made  for the men, not the men for the saturday", i.e. we  follow guidelines only if we gain advantage from them, not because they are valuable on its own, if you have any reason  why you want this in master let me know, thanks!

GiulioRomualdi:
Hi @traversaro. Devel is fine!

traversaro:
Just to clarify, now that I cleaned up a bit the release process of iDynTree, I think if we need a iDynTree 1.1 we can easily release it when we need it, without waiting years. : )

traversaro:
> 
> 
> Hi @traversaro. Devel is fine!

Perfect, then feel free to change the target branch.

GiulioRomualdi:
Rebased on top on devel

GiulioRomualdi:
The spurious commits are there because master is not merged in devel  
image

DanielePucci:
As for future reference, 
> "saturday that is made for the men, not the men for the saturday"

it is [Gospel of Mark, 2:27](https://www.biblegateway.com/passage/?search=Mark+2%3A27&version=NIV). I had to google for it, I believed it was actually Gospel of Mathew.

CC
@traversaro @GiulioRomualdi

traversaro:
> 
> 
> The spurious commits are there because master is not merged in devel
> image

Sorry, I fixed it now.

GiulioRomualdi:
> > The spurious commits are there because master is not merged in devel
> > image
> 
> Sorry, I fixed it now.

Rebased on devel. You can merge

traversaro:
Thanks!

S-Dafarra:
Sorry for the late comment. Was there a reason not to use an ``std::iterator``? I am not a great fan of using a pointer to double as iterator :thinking:

traversaro:
> 
> 
> Sorry for the late comment. Was there a reason not to use an `std::iterator`? I am not a great fan of using a pointer to double as iterator 🤔

What do you think @GiulioRomualdi ? Until we actually release iDynTree 1.1, we can change the interface any time we want, so feel free to propose a change if you see fit.

GiulioRomualdi:
[`std::iterator`](https://en.cppreference.com/w/cpp/iterator/iterator) has been deprecated in C++17, that's the reason why I didn't use it. I don't know which is the alternative :sad:

S-Dafarra:
To be honest I did not know. Apparently the alternative is simply to define your own iterator without inheriting from ``std::iterator``, but defining all the aliases (https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/).

PR #668, opened 2020-04-04
New matlab visualization

fjandrad:
This PR provides the functions to create a Matlab iDyntree visualizer.
It uses Matlab plotting and stl reading capabilities and the forward kinematics for floating base robots of iDyntree.

The two main functions are : 
- prepareVisualization
- updateVisualization 

A tutorial/testing script can also be found in [visualizeRobot](https://github.com/robotology/idyntree/blob/newMatlabVisualization/bindings/matlab/Visualization/visualizeRobot.m).

A new function was required to get better update timings. The function is 

````cpp 
std::vector<iDynTree::Matrix4x4> getWorldTransformsAsHomogeneous(std::vector<std::string> frameNames)`

The corresponding bindings and iDyntreeWrapper function were created.

### CODE REVIEW

File: bindings/iDynTree.i [hunk 1]
```diff
@@ -350,6 +353,8 @@ TEMPLATE_WRAP_MOTION_FORCE(ForceVector3, WRAP_FORCE, SET_NAME_FOR_WRAPPER,,)
 // High level interfaces
 %include "iDynTree/KinDynComputations.h"
 
+%template(MatrixVector) std::vector<iDynTree::Matrix4x4>;
```
traversaro, line 356: This is ambiguous, how we would call then a Vector of Matrix of different sizes?  I suggest to call  this `Matrix4x4Vector` .

bindings/matlab/Visualization/vis_iCub_test.m
```diff
@@ -0,0 +1,38 @@
+icubModelPath=[ getenv('ROBOTOLOGY_SUPERBUILD_SOURCE_DIR') '/build-linux/install/share/iCub/robots/iCubGazeboV2_5/'];
```
traversaro, line 1: Please avoid this kind of absolute references that could  change depending on the user specific setup. If this is not possible, feel free to remove the test.

bindings/matlab/Visualization/visualizeRobot.m
```diff
@@ -0,0 +1,56 @@
+% This script is an example on how to use the matlab Idyntree visualization
```
traversaro, line 1:
```suggestion
% This script is an example on how to use the matlab iDynTree visualization
```

bindings/matlab/Visualization/visualizeRobot.m
```diff
@@ -0,0 +1,56 @@
...
+
+% insert the path where the meshes may be found, it repaces what is require
+% d to complete the path inside the urdf visual mesh field.
+meshFilePrefix=[getenv('ROBOTOLOGY_SUPERBUILD_SOURCE_DIR') '/build-linux/install/share'];
```
traversaro, line 5: See https://github.com/robotology/idyntree/pull/668/files#r403447581 . If this file is meant to be used in practice, the meshFilePrefix should be some kind of external argument.

CHANGELOG.md
```diff
@@ -8,14 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added a new function to `iDyntreeWrappers` for the function `getWorldTransformsAsHomogeneous`.
```
traversaro, line 11:
```suggestion
- Added a new function to `iDynTreeWrappers` for the function `getWorldTransformsAsHomogeneous`.
```

doc/matlab_visualization.md
```diff
@@ -0,0 +1,43 @@
+# How to use Matlab Visualization
+The Matlab Visualization uses the Matlab capabilities for displaying meshes and figures and the floating-base kinematics of iDyntree to get the new poses of the links.
+It uses the [iDyntreewrappers](https://github.com/robotology/idyntree/tree/master/bindings/matlab/%2BiDynTreeWrappers) to handle the model and obtaining the required transforms.
```
traversaro, line 3: Avoid absolute links. You can just use relative links and they work fine as well.

doc/matlab_visualization.md
```diff
@@ -0,0 +1,43 @@
+# How to use Matlab Visualization
```
traversaro, line 1:
```suggestion
# How to use MATLAB Visualization
```
The correct capitalization of MATLAB, see the official documentation https://nl.mathworks.com/products/matlab.html .

src/high-level/include/iDynTree/KinDynComputations.h
```diff
@@ -378,6 +379,13 @@ class KinDynComputations {
      */
     iDynTree::Transform getWorldTransform(std::string frameName);
 
+    /**
+     * Return the transforms as a homogeneous matrices where the frame is the frame
+     * specified by frameIndex, and the reference frame is the world one
```
traversaro, line 384: What is `frameIndex` in this case?

src/high-level/src/KinDynComputations.cpp
```diff
@@ -1073,6 +1073,26 @@ Transform KinDynComputations::getWorldTransform(const FrameIndex frameIndex)
     return world_H_frame;
 }
 
+std::vector<iDynTree::Matrix4x4> KinDynComputations::getWorldTransformsAsHomogeneous(std::vector<std::string> frameNames)
+{
+    std::vector<iDynTree::Matrix4x4> worldTransforms;
+    int numberOfTransforms=frameNames.size();
+    for(int number=0;number<numberOfTransforms;number++)
```
traversaro, line 1080:
```suggestion
    for(int number=0; number<numberOfTransforms; number++)
```
please follow the stlye of the  rest of the code, thanks.

src/high-level/src/KinDynComputations.cpp
```diff
@@ -1073,6 +1073,26 @@ Transform KinDynComputations::getWorldTransform(const FrameIndex frameIndex)
...
+    {
+        std::string frameName=frameNames[number];
+        int frameIndex = getFrameIndex(frameName);
+        if( frameIndex < 0 )
+        {
+            worldTransforms.push_back(Transform::Identity().asHomogeneousTransform());
+        }
```
traversaro, line 1087: Silently failing is bad, because if a user puts a typo in the input frames, it should get a clear error, not the transform silently being returned as identity. If there is an error in the input data, I suggest to print a clear error and signal the error in the return value in some way (for example returning the empty vector could  be a way to signal the error).

bindings/matlab/Visualization/mesh2tri/mesh2tri.m
```diff
@@ -0,0 +1,96 @@
...
+        F(:,3)=repmat(IND,[4,1]);
+        Fe_I=[I(:),I(:)+1,I(:)+1,I(:)]; Fe_J=[J(:),J(:),J(:)+1,J(:)+1];
+        Fe = sub2ind(size(X),Fe_I,Fe_J);
+        Xe=mean(X(Fe),2); Ye=mean(Y(Fe),2);  Ze=mean(Z(Fe),2);
+        X=[X(:);Xe(:)]; Y=[Y(:);Ye(:)]; Z=[Z(:);Ze(:)];
+end
+
+V=[X(:),Y(:),Z(:)];
+
+end
```
traversaro, line 96: Including a BSD-licensed file is ok, but for tracking correctly the vendored dependencies it is important: 
* To place them under the existing `extern` directory 
* Add a link from where did you get the files

traversaro:
It would be useful if also @gabrielenava  could review this PR.

bindings/matlab/Visualization/plotMeshInWorld.m
```diff
@@ -0,0 +1,60 @@
...
+%
+% Copyright (C) 2019 Istituto Italiano di Tecnologia (IIT). All rights reserved.
+% This software may be modified and distributed under the terms of the
+% GNU Lesser General Public License v2.1 or any later version.
+
+        
+% Default values
+meshColor= [0.1843    0.3098    0.3098];
+transparency=0.5;
+scaleRatio=1/1000;
```
traversaro, line 20: The scale is not always 1/1000, it is saved in the `ExternalMesh::scale` attribute and should be used the valueu for which to scale.

bindings/matlab/Visualization/getMeshes.m
```diff
@@ -0,0 +1,108 @@
...
+    solidarray=iterator.next;
+    solids_number=size(solidarray,2);    
+    meshInfo=struct('meshFile',{},'mesh_triangles',{},'link_H_geom',{});
+    for solids=1:solids_number
+        if solidarray{solids}.isExternalMesh
+            meshName=split(solidarray{solids}.asExternalMesh.filename,':');
+            meshFile=meshName{2};
+            % Import an STL mesh, returning a PATCH-compatible face-vertex structure
+            mesh_triangles = stlread([meshFilePrefix meshFile]);
+            meshInfo(solids).meshFile=meshFile;
```
traversaro, line 44: URDFs (and so iDynTree::Model) support meshes also in collada (dae) format, and we use them a lot for example in  the new iRonCub model. Not supporting dae is ok, but it should be properly documented.

bindings/matlab/Visualization/getMeshes.m
```diff
@@ -0,0 +1,108 @@
...
+    linkMeshInfo(links).linkName=linkName;
+    solidarray=iterator.next;
+    solids_number=size(solidarray,2);    
+    meshInfo=struct('meshFile',{},'mesh_triangles',{},'link_H_geom',{});
+    for solids=1:solids_number
+        if solidarray{solids}.isExternalMesh
+            meshName=split(solidarray{solids}.asExternalMesh.filename,':');
+            meshFile=meshName{2};
+            % Import an STL mesh, returning a PATCH-compatible face-vertex structure
+            mesh_triangles = stlread([meshFilePrefix meshFile]);
```
traversaro, line 43: A URDF file can also contain absolute files (that is for example how the current visualizer works), so before appending a prefix you should check if the `filename` does not contain an absolute path.

traversaro:
Last request: could we change https://github.com/robotology/idyntree/blob/devel/bindings/iDynTree.i#L306 to `%template(LinksSolidshapesVector) std::vector< std::vector<iDynTree::SolidShape *>>;` .  All other types have the capital initial letter, and it would be good to be consistent.

bindings/matlab/Visualization/visualizeRobot.m
fjandrad: Is not meant to be used in practice is just like a tutorial or example on how to use the visualization.

src/high-level/src/KinDynComputations.cpp
fjandrad: I was thinking about that. I had done it in one of the functions I erased I think. I will add the reporterror. I was unsure if I want it to fail all together or just a warning.

bindings/matlab/Visualization/plotMeshInWorld.m
fjandrad: I was concerned about that cause I assume is a parameter that can be changed from the softy exporting the meshes. Although I think most CAD programs have mm as default unit.

bindings/matlab/Visualization/getMeshes.m
fjandrad: I'll have to check if Matlab loads far similarly to STL maybe both can be supported. I guess it will be needed to check the termination of the file.

bindings/matlab/Visualization/getMeshes.m
fjandrad: Can you put an example? That way I can see how to distinguish relative from absolute path. I  guess one way is to check if the file alone exists and if not use the prefix.

fjandrad:
> The first comment is: how is this supposed to be used? These files are not installed anywhere, so they are not found in the MATLAB path of the users that follow the documentation on how to use the bindings. Furthermore, could it make sense that instead of creating a new Visualization directory, we put this functions under the existing +iDynTreeWrappers directory so that installation is already sorted out?

That was something I had asked @gabrielenava in the beginning it makes sense to me to have it iDynTreeWrappers since I use the wrapper. So I will add them there and update the example. 

I do not know if it works in Octave. I would have to install and check. My gut feeling is that it doesn't. In the past I saw that Octave and Matlab differ when plotting and handling of figures cones in the picture. Do not know if that changed. If it's compatible it would be really nice.

fjandrad:
Also the visualize robot file should not go to the iDynTreeWrappers. Where should I put it? Is supposed to be an example/tutorial.

bindings/matlab/Visualization/getMeshes.m
traversaro: I am afraid that it will not be as easy as STL as collada is much more complex format (so I think just saying that we do not support it should  be enough).

bindings/matlab/Visualization/getMeshes.m
traversaro: See https://github.com/robotology/idyntree/issues/291 . I guess that if a file starts with `package://`  or `file://` for sure it is  not a relative path.

traversaro:
> I do not know if it works in Octave. I would have to install and check. My gut feeling is that it doesn't. In the past I saw that Octave and Matlab differ when plotting and handling of figures cones in the picture. Do not know if that changed. If it's compatible it would be really nice.

You can also just document that in Octave is not tested. 

> Also the visualize robot file should not go to the iDynTreeWrappers. Where should I put it? Is supposed to be an example/tutorial.

`examples/iDynTreeWrappers` seems ok, but let's hear @gabrielenava opinion. 

> There is also the file I used for benchmarking the times. Should I delete it? It served its purpose so now it does not really have a reason to stay in the repo. If we merge without squashing is enough for me to be there in the history.

Yes, any additional file in a repo is a burden to maintain, so if you do not have any use for it please delete it. You can also just copy it in an issue comment so if you need to recover it is there, even because ideally the commit history will need to be cleanup before merging the PR.

bindings/matlab/Visualization/getMeshes.m
fjandrad: Actually we are  tackling only the package case. Sibcce I do not have the resource finder. I'm directly asking for the path that would replace package. So in that way we are actually handling absolute paths and not relative paths.

bindings/matlab/Visualization/getMeshes.m
fjandrad: I found this https://it.mathworks.com/matlabcentral/fileexchange/45687-colladaparser I can give it a try. I will deal with the other changes and mention that for now we do not support saw. But maybe with this we might able to support it later. If it's ok with you I want to release this without DAE support and then if possible add it in another PR.

fjandrad: Of course MATLAB supports DAE in one or 2 of it's toolboxes...

bindings/matlab/Visualization/vis_iCub_test.m
fjandrad: I think that for avoiding that I would need to add a custom urdf with the meshes inside the testing or example folder

CHANGELOG.md
fjandrad: Done

src/high-level/include/iDynTree/KinDynComputations.h
fjandrad: Left over from copying something else.... changed to :

    

>  * Return the transforms as a homogeneous matrices where the frame is
>      * specified by name in the frameNames vector, and the reference frame is the world one
>      * (world_H_frame).

src/high-level/src/KinDynComputations.cpp
fjandrad: Now that I take another look at it,  getFrameIndex should show an error if the frame is not there. Is that not enough?

bindings/matlab/Visualization/mesh2tri/mesh2tri.m
fjandrad: Copied what you did for the MoxUnit, so it will only be installed if enable Matlab or generate matlab. Is that ok?

bindings/matlab/Visualization/visualizeRobot.m
fjandrad: moved to examples/iDynTreeWrappers , still need to figure out the meshes for the testing

bindings/matlab/Visualization/plotMeshInWorld.m
fjandrad: I see that it returns a vector size 3. Is there a case where the scale can be different for each axis? Is it why we need 3 values for the scale?

fjandrad: Done

src/high-level/src/KinDynComputations.cpp
S-Dafarra: I think it is still better to print something to have a sort of stack trace of the error.

bindings/matlab/Visualization/visualizeRobot.m
traversaro: You can also document that you need to manually set the prefix to the location where icub-models is install.

extern/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 #  iDynTree
 #  Copyright: (C) 2016 iCub Facility, Istituto Italiano di Tecnologia
-#  Authors: Silvio Traversaro <silvio.traversaro@iit.it>
+#  Authors: Silvio Traversaro <silvio.traversaro@iit.it>
```
traversaro, line 3:
```suggestion
#  Authors: Silvio Traversaro <silvio.traversaro@iit.it>
```

Please try to avoid adding  trailing whitespaces. You can configure most editors to automatically remove them.

bindings/matlab/Visualization/mesh2tri/mesh2tri.m
traversaro: Can you please point to where the `IDYNTREE_INTERNAL_MESH2TRI_PATH` variable is used and where the `mesh2tri.m` file is installed? I was not able to find it.

bindings/matlab/Visualization/getMeshes.m
traversaro: >  it's ok with you I want to release this without DAE support and then if possible add it in another PR.

Yes, not having DAE support is ok, as long as it is documented.

bindings/matlab/Visualization/getMeshes.m
traversaro: Yes, but in this line https://github.com/robotology/idyntree/pull/668/files#diff-34f42ee1b4d63ad197c2d3194a22bacfR42 you need to check if the initial part of the URI is `package://`, otherwise you are going to add the prefix also to absolute files (specified by `file://`).

bindings/matlab/Visualization/visualizeRobot.m
fjandrad: I wrote what the meshPrefix is and why is needed in the documentation I added in docs and inside the perpareVisualization and getMeshes functions.

bindings/matlab/Visualization/vis_iCub_test.m
traversaro: If  it is just an example, you can just document that this example depends on having icub-models installed, and you can have a variable that the user should replace with the installation prefix of icub-models on its machine.

bindings/matlab/Visualization/mesh2tri/mesh2tri.m
fjandrad: This is what I added in the extern folder:

```cpp
# mesh2tri ( https://it.mathworks.com/matlabcentral/fileexchange/28327-mesh2tri ) is a matlab function that allows converting a mesh format into a triangulation format.
# It is only installed if Matlab is enabled, not tested in Octave yet.
if (IDYNTREE_USES_MATLAB OR IDYNTREE_GENERATE_MATLAB) 
    # We save the location of mesh2tri in a variable that is then accessible to the test
    set(IDYNTREE_INTERNAL_MESH2TRI_PATH "${CMAKE_CURRENT_LIST_DIR}/mesh2tri/mesh2tri" PARENT_SCOPE)
endif()
```

But I think it is not correct. Could you point me into the right solution? I'm still not very proficient with CMakeLists...

src/high-level/src/KinDynComputations.cpp
fjandrad: Ok Then again I will probably repeat the same message but from the other function.

bindings/matlab/Visualization/mesh2tri/mesh2tri.m
fjandrad: It is used to convert from the mesh type of varaible that matlab generates for the sphere and cylinder to a triangulation to keep the same format as the stl files and the box geometries.
Code line is here:
https://github.com/robotology/idyntree/blob/cfac9c4847855b85a7377bbab2b4afe771c66f24/bindings/matlab/%2BiDynTreeWrappers/getMeshes.m#L109

src/high-level/src/KinDynComputations.cpp
S-Dafarra: Otherwise, you can run a check first with [``isFrameNameUsed()``](http://wiki.icub.org/codyco/dox/html/idyntree/html/classiDynTree_1_1Model.html#a6da12797ce7fd9bc747a6192e1a8632d) so that you avoid the other method to print the error before you.

bindings/matlab/Visualization/mesh2tri/mesh2tri.m
traversaro: I do not understand how this is working in your tests if you do not install `mesh2tri` . Where is this file supposed to be installed? I guess if you know its destination, you can just add a call to: 
~~~
install(FILES ${IDYNTREE_INTERNAL_MESH2TRI_PATH}/mesh2tri.m DESTINATION <whereYouWantToInstallIt>)
~~~

examples/iDynTreeWrappers/visualizeRobot.m
```diff
@@ -0,0 +1,93 @@
+% This script is an example on how to use the matlab iDynTree visualization
+
+% insert the path where the meshes may be found, it repaces what is require
+% d to complete the path inside the urdf visual mesh field.
+meshFilePrefix=[getenv('ROBOTOLOGY_SUPERBUILD_SOURCE_DIR') '/build-linux/install/share'];
+
+% create KinDynModel variable from another script
+icubModelPath=[ getenv('ROBOTOLOGY_SUPERBUILD_SOURCE_DIR') '/build-linux/install/share/iCub/robots/iCubGazeboV2_5/'];
+fileName='model.urdf';
```
traversaro, line 9: I  think it is more clear if you do something like that: 

~~~
% This example requires to use the URDF models installed by https://github.com/robotology/icub-models
% To do so, please substitute in the following the location of the install prefix of icub-models:
icubModelsInstallPrefix = "";
meshFilePrefix = [icubModelsInstallPrefix '/share']
icubModelPath = [icubModelsInstallPrefix '/share/iCub/robots//iCubGazeboV2_5/']
~~~

gabrielenava:
ok for me to review, but there are 234 changed files. To maximize my review efficiency it would be better if I focus on the files where my comment may be needed the most. On which flies should I focus @fjandrad ?

fjandrad:
> ok for me to review, but there are 234 changed files. To maximize my review efficiency it would be better if I focus on the files where my comment may be needed the most. On which flies should I focus @fjandrad ?

You see that amount because of the changes to the bindings, so focus on the changes inside iDynTreeWrappers and the folder I added  examples/iDyntreeWrappers

bindings/matlab/Visualization/getMeshes.m
fjandrad: Thanks, I was missing the piece of information between package and file, I added now a check for the keyword package and only add the mesh path in that case. Is there other specific words used? The only relevant reference I found with the `file://` notation is [this](https://answers.gazebosim.org//question/12387/referencing-models-from-other-packages-launch-files/) and it mentions also the `model://` option. Do you have some documentation where this keywords are mentioned?

bindings/matlab/Visualization/getMeshes.m
fjandrad: I created this issue https://github.com/robotology/idyntree/issues/670 for that, and will add documentation on missing .dae support for now

examples/iDynTreeWrappers/visualizeRobot.m
fjandrad: This would require the user to have iCubModels installed for the test no? I'm planning to add a urdf with the stl files inside the example folder so that it is self contained and runs out of the box. Is that fine for you?

bindings/matlab/Visualization/mesh2tri/mesh2tri.m
fjandrad: For my tests, I was adding the path manually. I was testing mainly the functionality and not the install process. I will test the install process now.

fjandrad: By default in many of the matlab repos that use bindings we add the install/mex folder so I will put the external visualization functions there directly. Should I put them inside a folder to signal they are for visualization or is fine just to add them there in the directory?

examples/iDynTreeWrappers/visualizeRobot.m
fjandrad: I also added the suggestion you put here

fjandrad:
Rebased on top of devel for the changes in CHANGELOG.md ,

bindings/matlab/Visualization/vis_iCub_test.m
fjandrad: This file was eliminated

bindings/matlab/Visualization/visualizeRobot.m
fjandrad: Done

fjandrad:
I managed to get a reasonable time optimization, see https://github.com/robotology/idyntree/issues/659.

I think is ready for its final reviews, I would not add anything else to this PR in terms of functionality. 
So feel free to review it. @traversaro @gabrielenava

bindings/iDynTree.i
```diff
@@ -302,8 +305,8 @@ TEMPLATE_WRAP_MOTION_FORCE(ForceVector3, WRAP_FORCE, SET_NAME_FOR_WRAPPER,,)
 
 %include "joints.i"
 
-%template(SolidshapesVector) std::vector<iDynTree::SolidShape*>;
-%template(linksSolidshapesVector) std::vector< std::vector<iDynTree::SolidShape *>>;
+%template(SolidShapesVector) std::vector<iDynTree::SolidShape*>;
+%template(linksSolidShapesVector) std::vector< std::vector<iDynTree::SolidShape *>>;
```
traversaro, line 309: As  I mentioned in another comment, it would be great to keep the `l` capital. 
```suggestion
%template(LinksSolidShapesVector) std::vector< std::vector<iDynTree::SolidShape *>>;
```

fjandrad: I'm becoming dislexic XD, I read the middle ones to be capital as in the original variables. Sure I'll change that.

traversaro:
> I managed to get a reasonable time optimization, see #659.
>
> I think is ready for its final reviews, I would not add anything else to this PR in terms of functionality.
So feel free to review it. @traversaro @gabrielenava

Great, as you may imagine reviewing a PR with 36 commits and 15K modified files is not obvious.  For keep the history of the repo sane, I suggest that you consolidete the commits in a limited number of commits, such as: 
* One commit for the modificaitons of the C++  and bindings stuff 
* One commit of the autogenerated code 
* One commit of the new iDynTreeWrappers stuff (so for example  @gabrielenava knows that he just need to read this commit)

As running `git rebase -i` across 36 commits and 300 files is not obvious, probably the easiest path is just to switch back to master branch and add again (as separate commits) the files that are changed. Probably it should be sufficient to go in the branch and do `git reset origin master`, that should leave all the files of the repo as locally modified  files.

bindings/iDynTree.i
fjandrad: Done @traversaro

fjandrad:
Should I reset on `master` or `devel`? This branch is rebased on devel already.

traversaro:
> 
> 
> Should I reset on `master` or `devel`? This branch is rebased on devel already.

Sorry it was a typo, I mean to  cleanup the commits w.r.t. ot the base branch of the PR.

bindings/matlab/matlab_mat4x4vec.i
```diff
@@ -0,0 +1,39 @@
...
+            for (int row = 0; row != nRows; ++row)
+            {
+                for (int col = 0; col != nCols; ++col)
+                {
+                    int index_sequence=(nCols*row + col);
+                    size_t overall_index=matrixN+ nMatrices*(row +nRows* col);
+                    d[overall_index]=content[index_sequence];
+                }
+            }
+            //free(content); //Tried to free the pointer but got error.
```
traversaro, line 29: Yes, the data is owned by the vector, so you should not try to delete it. This line can be deleted.
```suggestion
```

src/high-level/src/KinDynComputations.cpp
```diff
@@ -1073,6 +1073,27 @@ Transform KinDynComputations::getWorldTransform(const FrameIndex frameIndex)
...
+    std::vector<iDynTree::Matrix4x4> worldTransforms;
+    int numberOfTransforms=frameNames.size();
+    for(int number=0; number<numberOfTransforms; number++)
+    {
+        std::string frameName=frameNames[number];
+        int frameIndex = getFrameIndex(frameName);
+        if( frameIndex < 0 )
+        {
+            reportError("KinDynComputations","getWorldTransformsAsHomogeneous", "requested frameName not found in model");
+            worldTransforms.push_back(Transform::Identity().asHomogeneousTransform());
```
traversaro, line 1087: Sorry, that will not be enough, because people do not read the output of the code. If there is an error, and error should be reported in someway to the called, for example returning an empty vector.

src/high-level/src/KinDynComputations.cpp
```diff
@@ -1073,6 +1073,27 @@ Transform KinDynComputations::getWorldTransform(const FrameIndex frameIndex)
...
 
+std::vector<iDynTree::Matrix4x4> KinDynComputations::getWorldTransformsAsHomogeneous(const std::vector<std::string>& frameNames)
+{
+    std::vector<iDynTree::Matrix4x4> worldTransforms;
+    int numberOfTransforms=frameNames.size();
+    for(int number=0; number<numberOfTransforms; number++)
+    {
+        std::string frameName=frameNames[number];
+        int frameIndex = getFrameIndex(frameName);
+        if( frameIndex < 0 )
```
traversaro, line 1084: The correct check is frameIndex == iDynTree::FRAME_INVALID_INDEX:
```suggestion
        if (frameIndex  == iDynTree::FRAME_INVALID_INDEX)
```

examples/iDynTreeWrappers/visualizeRobot.m
```diff
@@ -0,0 +1,111 @@
...
+% Example: 
+% The test model has the following mesh file in the URDF
+% 'package://stl/sim_sea_2-5_root_link_prt-binary.stl'
+% The required file is inside the meshes folder in this directory,
+% so to complete the path we are missing meshes as a prefix to the full
+% path starting from this directory.
+meshFilePrefix='meshes/';
+% The model is in this same folder
+modelPath=[currentFolder '/'];
+fileName='model_uses_stl.urdf';
```
traversaro, line 16: Sorry, these meshes are too heavy to be loaded in the iDynTree repo. Can't we just keep the code that is in the  later comment?

bindings/matlab/matlab_mat4x4vec.i
fjandrad: OK thanks, because I was a bit worried about memory leakage.

src/high-level/src/KinDynComputations.cpp
fjandrad: Will do

examples/iDynTreeWrappers/visualizeRobot.m
fjandrad: Ok

src/high-level/src/KinDynComputations.cpp
fjandrad: Ok, I was following the logic implemented in other iDyntree Functions. Example : 
https://github.com/robotology/idyntree/blob/a92bdf9ca50ef5826034e190ba3f31f6e133e0ed/src/high-level/src/KinDynComputations.cpp#L892

Most of the getTransform functions have that logic, do you want me to change those as wel now that I'm at it?

fjandrad: Now this will happen when a wrong frame is requeste ( only this function though ):
image

DanielePucci:
CC
@claudia-lat @robotology/iit-dynamic-interaction-control

traversaro:
Thanks  @fjandrad , that is much better. @gabrielenava I think we need your review just on the modifications in the commit https://github.com/robotology/idyntree/pull/668/commits/8beaacc6f6d3cf1dfe359e1f3995817be762ff42 (and later minor fixes in https://github.com/robotology/idyntree/pull/668/commits/4f1678ad2efcce42dd551f1f1697dc707f4b8425 and https://github.com/robotology/idyntree/pull/668/commits/d664edc1b5923d0241cdaa709a8cc9d5e5ac716f).

traversaro:
@gabrielenava if you comment on the lines  from the PR instead of opening the separate commit and commenting on those, then the discussion are more manageable, because you can reply directly from the PR, i.e. this is good: 
comment_good

This is not convenient instead: 
comment_bad

bindings/matlab/+iDynTreeWrappers/prepareVisualization.m
```diff
@@ -0,0 +1,59 @@
...
+grid on;
+
+% Axis labels
+title('Robot Visualizer ');
+xlabel('{x}');
+ylabel('{y}');
+zlabel('{z}');
+
+% Apply views
+view(parent,custom_view);
```
gabrielenava, line 59: missing void line at the end also here (test to see if now my comment is fine)

traversaro:
To clarify, even in the "Files Changed" view of the PR, you can still select which commit (or files) to review, by changing the option under "Change from": 
specify_commits

bindings/matlab/+iDynTreeWrappers/updateVisualization.m
```diff
@@ -0,0 +1,27 @@
+function []=updateVisualization(KinDynModel,linkNames,transforms)
```
gabrielenava, line 1: I see there are not debug checks in the wrappers but maybe we can add them later (if necessary), based on the users feedback.

fjandrad: Which kind of debug things would you add?

fjandrad:
I added the lines at the end of the files and address your comments. @gabrielenava let me know if its fine for you.

bindings/matlab/+iDynTreeWrappers/updateVisualization.m
gabrielenava: For example, on the previous visualizer based on `irrlicht` sometimes the robot was not visualized correctly (I saw a white window), but the code was not stopping and the error message was only on the Matlab terminal or absent (e.g. I had this issue when using `softwareopengl` option in Matlab). I would say that we need to be sure that in case something in the visualization fails, it is clear for the user why the robot is not visualized correctly. But said this, it is too early now and  we need to first use the visualizer in order to understand which error messages may help the user.

traversaro:
Great, finally done. Can you please curate the commit once more so that we can merge? Thanks!

fjandrad:
@traversaro done I think is ready to merge or fast forward it should be on top of current devel

traversaro:
Thanks @fjandrad and @gabrielenava !

DanielePucci:
CC @robotology/iit-dynamic-interaction-control

Yeshasvitvs:
> A tutorial/testing script can also be found in visualizeRobot.

@fjandrad @traversaro the link in the first comments is broken, Please update it with the correct reference which, if I am not wrong is https://github.com/robotology/idyntree/blob/master/examples/iDynTreeWrappers/visualizeRobot.m

traversaro:
> 
> 
> > A tutorial/testing script can also be found in visualizeRobot.
> 
> @fjandrad @traversaro the link in the first comments is broken, Please update it with the correct reference which, if I am not wrong is https://github.com/robotology/idyntree/blob/master/examples/iDynTreeWrappers/visualizeRobot.m

Thanks, I updated it with the correct version, just linking the version v1.1.0 as the master version can eventually change in the future.

PR #676, opened 2020-04-21
Fix matlab bindings issue #675

lrapetti:


traversaro:
Thanks @lrapetti .  It is compiling fine now?

lrapetti:
> Thanks @lrapetti . It is compiling fine now?

yes, also @claudia-lat tested it

PR #724, opened 2020-08-14
feature: SHALLOW vs DEEP transform tree

elandini84:
#### SHALLOW vs DEEP transforms tree
* Now yarprobotstatepublisher can organize the robot's transforms as a shallow tree or as a deep one. This can be done by passing a new parameter to the yarprobotstatepublisher executable (tree-type). 
* A distinction between static transforms and variable ones is also performed

### DIFF
--- a/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
+++ b/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
@@ -21,6 +21,7 @@
 
 #include <iDynTree/ModelIO/ModelLoader.h>
 #include <iDynTree/KinDynComputations.h>
+#include <iDynTree/Model/Traversal.h>
 #include <iDynTree/yarp/YARPConversions.h>
 
 #include "robotstatepublisher.h"
@@ -78,6 +79,12 @@ bool YARPRobotStatePublisherModule::configure(ResourceFinder &rf)
 
     string modelFileName=rf.check("model",Value("model.urdf")).asString();
     m_period=rf.check("period",Value(0.010)).asDouble();
+    m_treeType=rf.check("tree-type", Value("SHALLOW")).asString();
+    if(m_treeType != "SHALLOW" && m_treeType != "DEEP")
+    {
+        yError("Wrong tree format. The only allowed values are \"SHALLOW\" or \"DEEP\"");
+        return false;
+    }
 
     Property pTransformclient_cfg;
     pTransformclient_cfg.put("device", "transformClient");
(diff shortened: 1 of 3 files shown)


src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
```diff
@@ -252,16 +254,95 @@ void YARPRobotStatePublisherModule::onRead(yarp::rosmsg::sensor_msgs::JointState
...
+                continue;
+
+            iDynTree::Transform base_H_frame = m_kinDynComp.getRelativeTransform(m_baseFrameIndex, frameIdx);
+            iDynTree::toYarp(base_H_frame.asHomogeneousTransform(), m_buf4x4);
+            m_iframetrans->setTransform(m_tfPrefix + model.getFrameName(frameIdx),
+                                        m_tfPrefix + model.getFrameName(m_baseFrameIndex),
+                                        m_buf4x4);
+        }
+    }
+    else
```
diegoferigo, line 272: It's quite common to make a mistake in the configuration file. As it is, mistakes are treated as `DEEP`, and it could be problematic if configuration files start being committed in downstream repos with errors.

Can you please put this branch in a `if (m_treeType == "DEEP")` and add an `else` that returns false notifying a configuration error?

src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
```diff
@@ -252,16 +254,95 @@ void YARPRobotStatePublisherModule::onRead(yarp::rosmsg::sensor_msgs::JointState
...
+
+        bool setOk = false;
+
+        //Processing joints instead of links since it's easier this way to distinguish between static transform and non static ones
+        for (size_t jointIndex=0; jointIndex < model.getNrOfJoints(); jointIndex++)
+        {
+            auto currJoint = model.getJoint(jointIndex);
+            iDynTree::LinkIndex parentLinkIndex = currJoint->getFirstAttachedLink();
+            iDynTree::LinkIndex linkIndex = currJoint->getSecondAttachedLink();
+            iDynTree::Transform parentLink_H_link = m_kinDynComp.getRelativeTransform(parentLinkIndex, linkIndex);
```
S-Dafarra, line 295: Notice that, depending on the choice of base, the parent link is not necessarily the first attached link (that's why the method is called ``getFirstAttachedLink`` and not ``getParentLink``). In order to get the parent and the child link of a given joint you should use [``getChildLinkIndexFromJointIndex``](https://robotology.github.io/idyntree/master/classiDynTree_1_1Traversal.html#af511b057b56553be5db56e34ca17c9a9) and [``getParentLinkIndexFromJointIndex``](https://robotology.github.io/idyntree/master/classiDynTree_1_1Traversal.html#a7c82fa00d69ac54fefdd89abcf1b2a5f) from ``Traversal``.

src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
```diff
@@ -78,6 +79,7 @@ bool YARPRobotStatePublisherModule::configure(ResourceFinder &rf)
 
     string modelFileName=rf.check("model",Value("model.urdf")).asString();
     m_period=rf.check("period",Value(0.010)).asDouble();
+    m_treeType=rf.check("tree-type", Value("SHALLOW")).asString();
```
diegoferigo, line 82: Referring to https://github.com/robotology/idyntree/pull/724#discussion_r471342972, maybe checking here that the value can be either SHALLOW or DEEP and return false already in the `configure` step would be even better.

S-Dafarra:
I am also a little confused about the meaning of the option. It seems to me that ``SHALLOW`` returns the transformations with respect to the base link while ``DEEP`` with respect to the parent link.
If this is correct:
- I think the options may be clearer if called "BASE" vs "RELATIVE", but that's a matter of tastes.
- The code could be simplified by iterating over the frames also in this case, IMHO.
If this is wrong:
- What is the need of expressing the transforms in different ways depending on the option? I would suggest documenting it a little bit more in the ``help`` call.

src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
elandini84: Thanks a lot for the comment. I'm in vacation now, but I'll get back to it as soon as possible.

randaz81:
I'm here commenting just the rationale behind this PR, i.e. 

> What is the need of expressing the transforms in different ways depending on the option? 
> I used this class a couple of year ago and I remember that ROS / RViz accept transforms both as relative to parent and as relative to base.

The 'DEEP' option improves the compatibility towards ROS, **which is unable to compute transforms between parallel branches of tree**.

Let's consider for example the TF tree or R1 (cer.urdf), **SHALLOW mode**, with `mobile_base_body_link `the root of tree (please click on the image to enlarge the tree)

yarprobotstatelauncher (1)

and **DEEP mode** (also obtained from ROS robotstatepublisher):
frames_ros

Both the trees are good for the purpose of visualizing the robot on RVIZ. But in the first case, you cannot ask ROS to return the transformation between two child nodes, e.g. the `camera_link` and the `hand_link`. (please note that `YARP tfClient`, instead, is capable of doing this computation). 
So the problem arises when you want to integrate a ROS module which needs to internally compute transforms between child nodes (e.g. typically all the SLAM modules, were the frame `world` is the root node)

src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp
randaz81: I agree

S-Dafarra:
Thanks for the explanation @randaz81! So, is the terminology DEEP / SHALLOW used also in ROS?

@elandini84 do you have any idea why the diff looks messed up? It seems that the whole file have been rewritten from scratch now.

S-Dafarra:
> > So, is the terminology DEEP / SHALLOW used also in ROS?
> 
> To clarify, I came up with the deep/shallow naming in a discussion with @elandini84/@randaz81 based on the shape of the resulting transform tree, so indeed the meaning of the option needs to be documented as it is far from being obvious. Furthermore, if anyone has a suggestion for a more intuitive name feel free to propose it.

Ah I see. If we look at the corresponding tree, by giving the transform w.r.t the parent, we obtain a tree with a lot of layers, each having a few elements, indeed a DEEP tree. If, instead, we give the transform w.r.t. the base we get a tree with all the links in a single layer, i.e. SHALLOW. I was not able to see this before, so yeah I guess some additional documentation may be helpful but I am ok with these names now.

Also, regarding
>  - The code could be simplified by iterating over the frames also in this case, IMHO.

@elandini84 already answered in the code with 
https://github.com/robotology/idyntree/blob/188f92c31a986aa13d393b48b010c03e6acf20f3/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp#L294

My only open, not blocking, point would be 
>@elandini84 do you have any idea why the diff looks messed up? It seems that the whole file has been rewritten from scratch now.

S-Dafarra (approved):

elandini84:
> Thanks for the explanation @randaz81! So, is the terminology DEEP / SHALLOW used also in ROS?
> 
> @elandini84 do you have any idea why the diff looks messed up? It seems that the whole file has been rewritten from scratch now.

I just got back from vacation and I just noticed the issue. I don't now why this happened... I'll try to redo the commit. Sorry for the inconvenience.

traversaro:
Hi @elandini84 , just to understand, did you address the comments from the reviewers?

elandini84:
> Hi @elandini84 , just to understand, did you address the comments from the reviewers?

Hi, I think so (but I'll report the links to code down here).
I added a check on the `tree-type` input parameter in the `configure` function

https://github.com/elandini84/idyntree/blob/fdd9eb07211a1397cf1f2798ad3a85b5aefdbcf7/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp#L83-L87

I then replaced the calls to `getFirstAttachedLink` and `getSecondAttachedLink` with ` getParentLinkIndexFromJointIndex` and `getChildLinkIndexFromJointIndex`

https://github.com/elandini84/idyntree/blob/fdd9eb07211a1397cf1f2798ad3a85b5aefdbcf7/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp#L297-L301

Regarding the comment posted by @diegoferigo 

> The PR looks overall good to me. What's puzzles me a bit are how static transforms are handled. I think that the main reason for the existence of static transform is to avoid streaming them all the time. However, this PR constantly streams them in the port callback, making their usage futile. Is there anything I am missing here?
I think that people from the AnDy team should have a look at this PR and maybe test it with their code since they are the only ones I know that they're using this tf bridge @Yeshasvitvs @lrapetti @kouroshD

I don't think I got it entirely. The `onRead` function avoids settings the static transforms multiple times by checking if they are already there. 

https://github.com/elandini84/idyntree/blob/fdd9eb07211a1397cf1f2798ad3a85b5aefdbcf7/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp#L303-L313

https://github.com/elandini84/idyntree/blob/fdd9eb07211a1397cf1f2798ad3a85b5aefdbcf7/src/tools/yarprobotstatepublisher/src/robotstatepublisher.cpp#L338-L345

And so, if I read the `FrameTransformClient` source code correctly, they are not transmitted. But maybe I got it wrong and/or this wasn't what he meant.
That said, since he tagged other people for their opinion, I thought I had to wait for their response.
If that's not the case, I apologize for taking so much time to subject my doubts.

diegoferigo (approved):
> I don't think I got it entirely. The onRead function avoids settings the static transforms multiple times by checking if they are already there.

Thanks for pointing out the precise location of the code, I missed this check when I first reviewed the PR. It addresses my comment, static transform are not streamed all the time :)

traversaro (approved):

traversaro:
As far as i can see, all comments have been addressed. @randaz81 it is ok to merge this for you? Thanks!

traversaro:
Merging, if there are any further changes feel free to propose a PR before the release, thanks.

PR #744, opened 2020-10-07
Optimize iDynTreeWrappers [WIP]

gabrielenava:
My story:

I am trying to use the [new iDynTree visualizer](https://github.com/robotology/idyntree/tree/master/bindings/matlab/%2BiDynTreeWrappers#matlab-native-visualization) with MATLAB. In my test, there is a free falling mass with initial upwards velocity and subject to a lateral force that falls down. Namely:

singleMass_11_45

 I wanted to profile my visualization function, because it seemed to me the visualizer was pretty slow. Here is the result of running MATLAB profiler:

Screenshot from 2020-10-07 11-34-32

so, it seems that actually the call to `setRobotState` is the one which is affecting the most the performances. In order to optimize the function, I initialized only once the iDynTree quantites (such as, e.g.,  iDynTree.Rotation()) by defining them as persistent variables. This  halves the time required by `setRobotState` to update the robot pose:

Screenshot from 2020-10-07 11-35-43

### DIFF
--- a/bindings/matlab/+iDynTreeWrappers/README.md
+++ b/bindings/matlab/+iDynTreeWrappers/README.md
@@ -1,4 +1,4 @@
-# idyntree high-level-wrappers
+# iDynTree high-level-wrappers
 
 A collection of Matlab/Octave functions that wraps the functionalities of (mainly) the iDyntree class `KinDynComputations`. For further information on the single wrapper, refer to the description included in each function.
 
@@ -73,6 +73,8 @@ The purpose of the wrapper is therefore to provide a simpler and easy-to-use int
 
 Not proper wrappers, they wrap more than one method of the class each. **Requirements:** compile iDyntree with Irrlicht `(IDYNTREE_USES_IRRLICHT = ON)`.
 
+**Warning**: the visualizer class may be deprecated in a future release.
+
 - [initializeVisualizer](initializeVisualizer.m)
 - [visualizerSetup](visualizerSetup.m)
 - [updateVisualizer](updateVisualizer.m)
@@ -80,9 +82,11 @@ Not proper wrappers, they wrap more than one method of the class each. **Require
 ## Matlab Native visualization
 
 Not actual wrappers, they use the iDynTreeWrappers in combination with the MATLAB patch plotting functions to visualize the robot.
+
 **Disclaimers**:
-- This visualization **has not been tested** with Octave.
-- At the moment, there is **no support** for .dae mesh files.
+
+1) This visualization **has not been tested** with Octave.
+2) At the moment, there is **no support** for `.dae` mesh file format.
 
 - [prepareVisualization](prepareVisualization.m)
 - [updateVisualization](updateVisualization.m)
--- a/bindings/matlab/+iDynTreeWrappers/getMeshes.m
+++ b/bindings/matlab/+iDynTreeWrappers/getMeshes.m
@@ -1,118 +1,147 @@
-function [linkMeshInfo,map]=getMeshes(model,meshFilePrefix)
-% We use the iDyntree information to obtain the files containing the meshes
-% and we link them to their link names
-% Gets the mesh information for each link in the model.
-%     - Iputs:
-%         - `model` : iDyntree model loaded from a URDF.
-%         - `meshFilePrefix` : Path in which we can find the meshes. As an example the path to the mesh in a iCub urdf is `'package://iCub/meshes/simmechanics/sim_sea_2-5_root_link_prt-binary.stl'
-% `. `meshFilePrefix` should replace package to allow to find the rest of the path.
-%     - Outputs:
-%         - `map`        : Cell array having both the names of the meshes and the associated link
-%         - `linkMeshInfo` : Struct array that contain the link name and a struct (`meshInfo`) that contains the name of file or if is a simple geometry, the triangulation ( edges and vertices of the mesh ) and the link to geometry transform.
-%
-% NOTE: at the moment only STL files are supported.
-%
-% Author : Francisco Andrade (franciscojavier.andradechavez@iit.it)
-%
-% Copyright (C) 2019 Istituto Italiano di Tecnologia (IIT). All rights reserved.
-% This software may be modified and distributed under the terms of the
-% GNU Lesser General Public License v2.1 or any later version.
+function [linkMeshInfo, map] = getMeshes(model, meshFilePrefix)
 
+    % GETMESHES uses the iDynTree information to obtain the files containing 
+    %           the meshes and links them to their link names.
+    %
+    % FORMAT: [linkMeshInfo, map] = getMeshes(model, meshFilePrefix)
+    %
+    % INPUTS:  - model: iDyntree model loaded from a URDF.
+    %          - meshFilePrefix: Path in which we can find the meshes. As an example 
+    %                           the path to the meshes in an iCub urdf is 
+    %                           'package://iCub/meshes/simmechanics/sim_sea_2-5_root_link_prt-binary.stl'.
+    %                           'meshFilePrefix' should replace 'package' to 
+    %                            allow to find the rest of the path. 
+    %
+    % OUTPUTS: - map: Cell array having both the names of the meshes and the associated link
+    %          - linkMeshInfo: Struct array that contain the link name and a struct (`meshInfo`) 
+    %                          that contains the name of file or if is a simple geometry, the 
+    %                          triangulation (edges and vertices of the mesh) and the link to 
+    %                          geometry transform.
+    %
+    % Author : Francisco Andrade (franciscojavier.andradechavez@iit.it)
+    %          Modified by: Gabriele Nava (gabriele.nava@iit.it)
+    %    
+    % Copyright (C) 2019 Istituto Italiano di Tecnologia (IIT). All rights reserved.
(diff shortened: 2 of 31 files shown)


traversaro:
Cool! Just to let you know, I think another improvement could be to use the fromMatlab functions to convert MATLAB vectors to iDynTree one, instead of using a manual for loop that invokes the `setVal` method, however this can be done in a future PR.

traversaro (approved):

traversaro:
Nitpick (i.e. feel free to ignore): I wonder if instead of using persistent variables, we could just add this buffers in `KinDynModel` to avoid multithreading problems in the future, if two threads in the same MATLAB interpreter (I do not even know if it is possible) call setRobotState for two different `KinDynModel` .

traversaro:
The Octave unit tests are failing in a bad way, can you check if the units tests in Matlab are running as expected?

gabrielenava:
>The Octave unit tests are failing in a bad way, can you check if the units tests in Matlab are running as expected?

I am having some issues also in MATLAB: in case we simulate, one after the other, two robots with different number of DoFs, it is required to run `clear all` otherwise `setRobotState` will exit with an error, despite the `KinDynModel` has been deleted from workspace. 

I don't like this because `clear all` is not recommended by MATLAB as it decreases code performances. Also, I don't like the fact that the vector size is stored despite all variables are deleted from workspace, it is something confusing.

I need to think more about this performances improvement.

traversaro:
> maybe is related to this comment?
> 
> > Nitpick (i.e. feel free to ignore): I wonder if instead of using persistent variables, we could just add this buffers in KinDynModel to avoid multithreading problems in the future, if two threads in the same MATLAB interpreter (I do not even know if it is possible) call setRobotState for two different KinDynModel .

Yes, I think that having a private buffers attribute of that structure to store all the buffers used by the implementation would probably solve all the problems.

gabrielenava:
>Yes, I think that having a private buffers attribute of that structure to store all the buffers used by the implementation would probably solve all the problems.

I followed the suggestion, and with [this commit]() I've done the following:

- created a buffer inside `KinDynModel`. The wrapper is  created once inside [loadReducedModel](https://github.com/robotology/idyntree/pull/744/files#diff-7114c10e68cd3c12f9d1676cccbf7578R64) and initializes all quantities required by the other wrappers;

- fixed a bug in `getModelVel` according to its description, it should return both the floating base and the joints velocities. It was however initializing a vector with dimension NDOF and not NDOF+6: (see [here](https://github.com/robotology/idyntree/pull/744/files#)). 

- transformed an [error](https://github.com/robotology/idyntree/pull/744/files#diff-e6571d6f1c4b72b9a3442ba7f34b3b82R33) into warning in `getNrOfDegreesOfFreedom` (otherwise there is an unnecessary error when a model with no joints (e.g. a single mass) is loaded and DEBUG mode is activated).

- cleaned readme and functions descriptions;

- I wanted to better understand the new iDynTree visualizer, therefore I went through all the functions in order to get a better idea about how they work. In the process, I cleaned indentation, typos,  functions descriptions and so on because in this way it was easier for me to follow the code. I then committed also my cleanup.

Finally, I run ctest on my laptop for both MATLAB and Octave: it exited without errors.

Now I will run a [testWrapper](https://github.com/gabrielenava/mbs_app/blob/master/test-idyntreeWrappers/src/wrappersTest.m) that I have in my repo, and with the MATLAB profile I will compare the performances with the previous implementation to check the effectiveness of the optimization.

PS: I also tested the old iDynTree visualizer that worked fine.

gabrielenava:
In the end I repeated the test in https://github.com/robotology/idyntree/pull/744#issue-499134380:

test

Time employed by `setRobotState` before optimization:

Screenshot from 2020-10-09 14-14-43

Time employed after:

Screenshot from 2020-10-09 14-09-55

We still halved the time also with this solution, and Octave is not failing anymore.

traversaro:
Note that avoiding all the setVal calls (such as https://github.com/robotology/idyntree/pull/744/files#diff-884723fb6e7c8456b2b6b8f9951e215eL207) substiuting them with fromMatlab can help you in reducing also the number of `SwigRef.subref` calls.

bindings/matlab/+iDynTreeWrappers/getMeshes.m
```diff
@@ -1,118 +1,147 @@
...
-                box_dimensions(2)=solidarray{solids}.asBox.y;
-                box_dimensions(3)=solidarray{solids}.asBox.z;
-                mesh_triangles=calculateMeshFromBox(box_dimensions);
-            end
-            if solidarray{solids}.isSphere
-                meshInfo(solids).meshFile='sphere';
-                radius=solidarray{solids}.asSphere.radius;
-                mesh_triangles=calculateMeshFromSphere(radius);
+                meshInfo(solids).scale    = [1,1,1];
+
```
traversaro, line 74: Nitpick: Please avoid commiting trailing whitespaces, as they can create problems in comparing code, creating spurious whitespace only changes that to not correspond to real changes (see https://github.com/robotology/idyntree/pull/744/files#diff-eddfdc7bb22a31340c851c77734e51c6L27). Most editor have options to automatically avoid to commit trailing whitespaces, but after a quick search I am afraid that MATLAB does not.

traversaro:
> In the process, I cleaned indentation, typos, functions descriptions and so on because in this way it was easier for me to follow the code. I then committed also my cleanup.

For the future, please try to keep actual changes to the code and cleanup at least in separate commits, for two reasons: 
* A commit that changes ~800 lines of code of which only 10/20 lines actually change the logic is much more difficult to review then a change of 10/20 lines, then followed by a cleanup of 800 lines. 
* This change is done in `master` and it make sense to do so, as it contains bug fixes. However, being done in master, it will need to be forward merged to `devel`: heavily changing the indentation of the code means that if any of that code was modified also in devel, it is not possible to automatically merge the changes,  but rather the changes needed to be manually merged. Fortunately this time there are no changes in those classes in devel.

For more details on this, see https://google.github.io/eng-practices/review/developer/small-cls.html, in particular the section "Separate Out Refactorings". However, for this time  it is not a big problem.

traversaro:
Can you drop a line in https://github.com/robotology/idyntree/blob/master/CHANGELOG.md#changed mentioning the improved performance of MATLAB's iDynTreeWrappers ? Thanks!

traversaro:
You followed some style guide or automatic tool to cleanup the MATLAB files? If you did so, it probably make sense to mention these somewhere so that we have a reference for this, and we don't have everyone that changes the MATLAB files style based on its own taste.

traversaro:
By the way, if you want to deprecate a function, I suggest to mention in the CHANGELOG that it has been deprecated  .

bindings/matlab/+iDynTreeWrappers/README.md
```diff
@@ -73,16 +73,20 @@ The purpose of the wrapper is therefore to provide a simpler and easy-to-use int
 
 Not proper wrappers, they wrap more than one method of the class each. **Requirements:** compile iDyntree with Irrlicht `(IDYNTREE_USES_IRRLICHT = ON)`.
 
+**Warning**: the visualizer class may be deprecated in a future release.
```
traversaro, line 76: According to the documentation of this functions, they are being deprecated, so I would avoid the `may be`. However, as far as I remember back in time with @fjandrad we decided not to deprecate this functions as th new visualizer was not working in Octave.

traversaro:
Hi @gabrielenava do you plan to work on this? I wanted to relase a 1.2 version of iDynTree, if you plan to work on this in the next days I can wait, otherwise I will go forward and this can go in a following release.

gabrielenava:
I planned to work on this, if possible I would wait until Friday: if I make it we can add it to the release, otherwise feel free to proceed.

traversaro:
> 
> 
> I planned to work on this, if possible I would wait until Friday: if I make it we can add it to the release, otherwise feel free to proceed.

Deal!

gabrielenava:
cc @nunoguedelha @CarlottaSartore

traversaro (changes requested):
Marking as "Requested Changes" explicitly so that it is clear why this is not being merged.

gabrielenava:
make sense to re-apply the changes (for what concerns the MATLAB visualizer) on top of #784

traversaro:
To avoid confusion in the repo, let's close this issue. Whenever you want to work on it, feel free to re-open it. This may be interesting also for other visualizer users @lrapetti @nunoguedelha  .

nunoguedelha:
So, @gabrielenava @traversaro , the work left to do would be:
- replace the `setVal` function calls.
- address the unresolved review comments.
- rebase on top of #784.

Am I missing something?
Is anyone working on this?

I don't think I'll have time to work on this before the 3rd March, but I after that, I can take the task.

gabrielenava:
>Am I missing something?

I think that's it, I am not working on it because of time constraints. Feel free to take the task after 3rd of March :-)

traversaro:
> address the unresolved review comments.

Yes, to clarify the unresolved comments are:
 * Removing trailing whitespace, if there are any left (https://github.com/robotology/idyntree/pull/744#discussion_r502484195)
 * Update the CHANGELOG ( https://github.com/robotology/idyntree/pull/744#issuecomment-706230376 )
 * Mention the way in which MATLAB code style has been changed 

The `SetVal` is yet an additional improvement that can also go in a following PR.

PR #802, opened 2021-01-22
Feature/visualization++

S-Dafarra:
It does the following:
- It adds a camera animator to move the camera via the mouse
- It fixes the aspect ratio to avoid strange effects when resizing
- It adds an interface to add custom frames to the visualization
- It fixes the floor grid colors
- It adds an interface to have additional textures (helpful if you need to render to a different size with a different environment and to get the pixels of the corresponding image)
- It fixes the visualizations of ``STL`` meshes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@
 
 cmake_minimum_required(VERSION 3.16)
 
-project(iDynTree VERSION 2.0.2
+project(iDynTree VERSION 2.0.3
                  LANGUAGES C CXX)
 
 # Disable in source build, unless Eclipse is used
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@
 
 cmake_minimum_required(VERSION 3.16)
 
-project(iDynTree VERSION 2.0.3
+project(iDynTree VERSION 2.0.100
                  LANGUAGES C CXX)
 
 # Disable in source build, unless Eclipse is used
--- a/src/visualization/include/iDynTree/Visualizer.h
+++ b/src/visualization/include/iDynTree/Visualizer.h
@@ -35,6 +35,8 @@ public:
 
     /**
      * Enable the control of camera via the mouse.
+     * By default, left button rotates the camera, the right button translates it on the plane.
+     * The middle button, translates the camera up and down, while the wheel increases or decreases the zoom.
      */
     virtual void enableMouseControl(bool enabled = true) = 0;
 
@@ -576,6 +578,12 @@ public:
     virtual Transform getWorldLinkTransform(const LinkIndex& linkIndex) = 0;
 };
 
+/**
+ * The interface for an object that can be used as an additional target for the renderer.
+ * This allows rendering the scene using dimensions and environment that are different from
+ * the main window. The camera is in common. Any camera change in the main window is also
+ * reflected in the other textures.
+ */
 class ITexture
 {
 public:
@@ -604,7 +612,11 @@ public:
      * @brief Get the pixels of the texture.
      *
      * Remember to call draw() first.
-     * @param pixels The output pixels
+     * @param pixels The output pixels. The size of the vector will be equal to the total number of
+     * pixels of the rendered texture, i.e. width of the texture times its height. Both these two parameters
+     * can be set in the textureOptions passed to the method ITexturesHandler::add. The pixels are saved in
+     * col-major format.
+     * @note This operation may affect the time performances of the visualizer, especially if the texture is large.
      * @return True in case of success, false otherwise
      */
     virtual bool getPixels(std::vector<PixelViz>& pixels) const = 0;
--- a/src/visualization/include/iDynTree/Visualizer.h
+++ b/src/visualization/include/iDynTree/Visualizer.h
@@ -100,6 +100,7 @@ public:
     /**
      * Get a pointer to the CameraAnimator object.
      * It is not supposed to be deallocated, and its lifespan coincides with the one of the ICamera
+     * It is accessible only after the visualizer has been initialized.
      */
     virtual ICameraAnimator* animator() = 0;
 
--- a/src/visualization/include/iDynTree/Visualizer.h
+++ b/src/visualization/include/iDynTree/Visualizer.h
@@ -653,6 +653,11 @@ public:
     IVectorsVisualization& vectors();
 
     /**
+     * Get a reference to the internal IFrameVisualization interface.
+     */
+    IFrameVisualization& frames();
+
+    /**
      * Wrap the run method of the Irrlicht device.
      */
     bool run();
--- a/src/visualization/include/iDynTree/Visualizer.h
+++ b/src/visualization/include/iDynTree/Visualizer.h
@@ -239,6 +239,11 @@ public:
     virtual void setBackgroundColor(const ColorViz & backgroundColor) = 0;
 
     /**
+     * Set the floor grid color.
+     */
+    virtual void setFloorGridColor(const ColorViz & floorGridColor) = 0;
+
+    /**
      * Set the ambient light of the enviroment.
      */
     virtual void setAmbientLight(const ColorViz & ambientLight) = 0;
@@ -546,16 +551,10 @@ struct VisualizerOptions
      */
     double rootFrameArrowsDimension;
 
-    /**
-     * The bacground color
-     */
-    ColorViz backgroundColor;
-
     VisualizerOptions(): verbose(false),
                          winWidth(800),
                          winHeight(600),
-                         rootFrameArrowsDimension(1.0),
-                         backgroundColor(0.0,0.4,0.4,1.0)
+                         rootFrameArrowsDimension(1.0)
     {
     }
 };
@@ -642,6 +641,11 @@ public:
      * Return an interface to manipulate the visualization environment.
      */
     IEnvironment& enviroment();
+
+    /**
+     * Return an interface to manipulate the environment in the additional texture.
+     */
+    IEnvironment& textureEnviroment();
 
     /**
      * Get a reference to the internal IVectorsVisualization interface.
--- a/src/visualization/include/iDynTree/Visualizer.h
+++ b/src/visualization/include/iDynTree/Visualizer.h
@@ -694,14 +694,7 @@ public:
      *
      * \note this is called implicitly when addModel is called for the first time.
      */
-    bool init(const VisualizerOptions = VisualizerOptions());
-
-    /**
-     * Initialize the visualization.
-     *
-     * \note this is called implicitly when addModel is called for the first time.
-     */
-    bool init(const VisualizerOptions& visualizerOptions, const VisualizerOptions& textureOptions);
+    bool init(const VisualizerOptions& visualizerOptions = VisualizerOptions());
 
     /**
      * Get number of models visualized.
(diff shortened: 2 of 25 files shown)


src/visualization/src/Texture.cpp
```diff
@@ -0,0 +1,116 @@
...
+                pixelIrrlicht = irr::video::SColor(*(unsigned int*)(buffer + (height * pitch) + (width * bytes)));
+                pixels[i].width = width;
+                pixels[i].height = height;
+                pixels[i].r = pixelIrrlicht.getRed();
+                pixels[i].g = pixelIrrlicht.getGreen();
+                pixels[i].b = pixelIrrlicht.getBlue();
+                pixels[i].a = pixelIrrlicht.getAlpha();
+                ++i;
+            }
+        }
```
S-Dafarra, line 110: FYI, this is pretty inefficient. I would have liked to output directly ``buffer``, but it should have been necessary to implement also the lock/unlock mechanism required by ``irrlicht``. In addition, it was not clear how to pass the information on how to interpret that buffer. For RGB images, having the raw buffer may be handy with ``YARP`` images, thus avoiding two double for loops. In fact, ``irrlicht`` and ``YARP`` seem to store RGB images in the same way. It is not the same for RGBA. In fact, ``irrlicht`` uses ARGB instead.
Bottom line is, I did not find a better way to output the raw image. In any case, this has an effect only when the corresponding texture is large, or when getting the pixels of many textures.

CMakeLists.txt
```diff
@@ -8,7 +8,7 @@
 
 cmake_minimum_required(VERSION 3.16)
 
-project(iDynTree VERSION 2.0.2
+project(iDynTree VERSION 2.0.3
```
traversaro, line 11: Feel free to bump more (like 2.0.100), as 2.0.3 in the future could be be a patch release on the master branch.

src/visualization/include/iDynTree/Visualizer.h
```diff
@@ -466,6 +576,41 @@ class IModelVisualization
     virtual Transform getWorldLinkTransform(const LinkIndex& linkIndex) = 0;
 };
 
+class ITexture
```
traversaro, line 579: Can you add here or in some other related interface what you mean by "Texture"? In 3D Graphics texture is typically used to refer to 2D images that is then mapped to the surface of a 3D model to describe its appearance (see https://en.wikipedia.org/wiki/Texture_mapping). In this context instead if I understood correctly the texture is a sort of "render target" (see the second meaning in https://www.khronos.org/opengl/wiki/Texture). As this could be confusing for users, could you clarify this? Thanks!

src/visualization/src/IrrlichtUtils.h
```diff
@@ -73,7 +66,25 @@ inline iDynTree::Rotation irr2idyntree_rot(const irr::core::vector3df & rotIrr)
 
 inline const irr::core::vector3df idyntree2irr_rot(const iDynTree::Rotation & rot)
 {
-    return idyntree2irr_rpy(rot.asRPY());
+    irr::core::matrix4 irrTransform(irr::core::matrix4::EM4CONST_IDENTITY);
+    //Transposing the matrix since irrlicht uses left handed frames
```
traversaro, line 70: Can you elaborate a bit more on this? I thought that the rendering problem of STL meshes was solved by changing the scale of the X variable, instead also this is necessary? Did you tested this with the iCub 2.5 DAE models after this changes?

src/visualization/include/iDynTree/Visualizer.h
```diff
@@ -466,6 +576,41 @@ class IModelVisualization
...
+     */
+    virtual ColorViz getPixelColor(unsigned int width, unsigned int height) const = 0;
+
+    /**
+     * @brief Get the pixels of the texture.
+     *
+     * Remember to call draw() first.
+     * @param pixels The output pixels
+     * @return True in case of success, false otherwise
+     */
```
traversaro, line 609: Can you clarify here: 
* the size of std::vector<PixelViz>
* if the pixel are stored in row major or col major order?

Now that this docs seems to be duplicated also in the concrete Texture implementation, so either update also that one, or remove it.

traversaro (changes requested):
Some comments, but overall it seems great, thanks!

traversaro:
By the way, this new features (especially the mouse control and the frames) could be used to enhance idyntree-model-view ( https://github.com/robotology/idyntree/blob/master/src/tools/idyntree-model-view.cpp ) and make it a quick and dirty tool for easy inspection of URDF models frame placement. fyi @Nicogene

src/visualization/include/iDynTree/Visualizer.h
S-Dafarra: Yes, the correct interpretation would be that the texture is a different target for the renderer. Do you think that adding some documentation may be sufficient, or you about a better naming?

traversaro: As you prefer, I think that even just documentation is ok.

CMakeLists.txt
S-Dafarra: Done in https://github.com/robotology/idyntree/pull/802/commits/b10c4b8fbe6c016a574fac6c4ca1738680e85c7a

src/visualization/include/iDynTree/Visualizer.h
S-Dafarra: I have updated the documentation in https://github.com/robotology/idyntree/pull/802/commits/b7267fc3ba69363f2fcadd9fc53f37e1dd845d7f

src/visualization/src/IrrlichtUtils.h
S-Dafarra: I reverted the changes in https://github.com/robotology/idyntree/pull/802/commits/0a6f821fed94b0ae34a1ce8c747c33fa5a06b116. 
Regarding the visualization, I run ``idyntree-model-view`` with:
- the `iCubGenova09`` urdf
image
- the ``iCubGenova04`` urdf
image
- a custom ``iCubGenova04`` urdf with dae meshes
image

src/visualization/include/iDynTree/Visualizer.h
S-Dafarra: I have updated the docs in https://github.com/robotology/idyntree/pull/802/commits/b7267fc3ba69363f2fcadd9fc53f37e1dd845d7f and removed the duplicated documentation in https://github.com/robotology/idyntree/pull/802/commits/d337204fa745b9b28cac986727a5f3d7bc5c9fd8.

S-Dafarra:
> By the way, this new features (especially the mouse control and the frames) could be used to enhance idyntree-model-view ( https://github.com/robotology/idyntree/blob/master/src/tools/idyntree-model-view.cpp ) and make it a quick and dirty tool for easy inspection of URDF models frame placement. fyi @Nicogene

I enabled the mouse control for ``idyntree-model-view`` in https://github.com/robotology/idyntree/pull/802/commits/2e71c6fbc9a7fbb75068e2cebe558678325e9963.

src/visualization/src/FrameVisualization.h
```diff
@@ -0,0 +1,71 @@
...
+
+        FrameVisualization(const FrameVisualization& other) = delete;
+
+        FrameVisualization& operator=(const FrameVisualization& other) = delete;
+
+        /**
+         * @brief Add a frame in the visualization
+         * @return The frame index.
+         */
+        virtual size_t addFrame(const Transform& transformation, double arrowLength = 1.0) final;
```
prashanthr05, line 50: Maybe it would also be convenient to have a method that removes a frame?

src/visualization/src/IrrlichtUtils.h
traversaro: Ok thanks! However, if I recall correctly is that the problem with the stl meshes on iCub3 (the one that we imagine was fixed by the STL load fix) only appeared when the legs or arm were moved from the zero position, so after you reverted the change to the rotation did you check if the visualization is still working as expected?

S-Dafarra: Here the result when reading the joints from the control boards:
- ``iCubGenova09`` urdf
image
- ``iCubGenova04`` urdf
image
- ``iCubGenova04`` urdf with custom DAE meshes
image

src/visualization/src/FrameVisualization.h
S-Dafarra: I have tried adding the method to remove a frame, but the logic behind was not the best. Since I am returning an index as an identifier, I do not like what happens when you remove a frame. Should that index be reused or not? Also, with the ``getNumberOfFrames`` method you can understand whether an index is valid or not. But if you remove one frame, this does not hold anymore as the total number of frames will be lower or equal than the maximum index. Hence, for the time being, I preferred to allow controlling the visibility of a frame with  https://github.com/robotology/idyntree/pull/802/commits/98e804a2272868576bd5f3eee0ac51c39a547429. The net result is the same and the logic behind remains straightforward.

src/visualization/src/IrrlichtUtils.h
traversaro: Great!

traversaro (approved):

traversaro:
Ok for me, waiting for @prashanthr05 approval as well. 

@S-Dafarra it is ok for me to Squash and Merge or you want to revise the commits?

S-Dafarra:
> Ok for me, waiting for @prashanthr05 approval as well.
> 
> @S-Dafarra it is ok for me to Squash and Merge or you want to revise the commits?

It is ok to squash for me!

src/visualization/src/FrameVisualization.h
prashanthr05: Great!

src/visualization/src/CameraAnimator.cpp
```diff
@@ -0,0 +1,332 @@
...
+    initialTransformation(0,1) = xAxis.Y;
+    initialTransformation(0,2) = xAxis.Z;
+
+    initialTransformation(1,0) = yAxis.X;
+    initialTransformation(1,1) = yAxis.Y;
+    initialTransformation(1,2) = yAxis.Z;
+
+    initialTransformation(2,0) = zAxis.X;
+    initialTransformation(2,1) = zAxis.Y;
+    initialTransformation(2,2) = zAxis.Z;
```
prashanthr05, line 183: Could you just document this row-major order stacking for the axis vectors and why we use this ordering and the difference from column-major ordering? (rotation and its inverse)

prashanthr05 (approved):
In general, ok for me. I just had one clarification request in the CameraAnimator class.

src/visualization/src/CameraAnimator.cpp
S-Dafarra: I have added the relevant comment in https://github.com/robotology/idyntree/pull/802/commits/e20efdaca58282ce3c16895e871fd369b84e50ce

https://github.com/robotology/idyntree/blob/e20efdaca58282ce3c16895e871fd369b84e50ce/src/visualization/src/CameraAnimator.cpp#L172-L180

prashanthr05: Thank you. Very clear!

S-Dafarra: Indeed, that transpose caused me a lot of thinking in the first place. Initially, it was not transposed, but the visualization was clearly wrong. So I thought I was doing some mistakes in the interpretation, I tried the transpose and everything went fine. Later on, when I was debugging a problem with the ``stl`` meshes being in the wrong position, I discovered this thing about ``irrlicht`` using left-handed frames, and then everything made sense :sweat_smile:

PR #831, opened 2021-02-15
Introduce color palette in the Visualizer

GiulioRomualdi:
This PR introduces the color palette in the visualizer.
I implemented only two palettes `vanilla` and `meshcat`. The `vanilla` palette is the original one and it is loaded by default. 

I also modified the `idyntree-model-view` application to load different palettes (you can pass the argument with `-c`)

https://user-images.githubusercontent.com/16744101/107950983-d89f5e80-6f97-11eb-8f05-2b0fcbecdc23.mp4

The only two problems that we have right now are:
1. the model is really dark (is the sun correctly placed?) 
2. the camera is not well oriented at the startup (at least for `iCubGazeboV2_5` model)

cc @S-Dafarra @traversaro @Giulero

### DIFF
--- a/src/visualization/src/Visualizer.cpp
+++ b/src/visualization/src/Visualizer.cpp
@@ -148,6 +148,34 @@ struct Visualizer::VisualizerPimpl
      */
     TexturesHandler m_textures;
 
+    double rootFrameArrowsDimension;
+    struct ColorPalette
+    {
+        irr::video::SColorf background;
+        irr::video::SColor  gridColor;
+
+        irr::video::SColor xAxis;
+        irr::video::SColor yAxis;
+        irr::video::SColor zAxis;
+    };
+    std::unordered_map<std::string, ColorPalette> m_palette;
+
+    void initializePalette()
+    {
+        irr::u32 alphaLev = 20;
+        // vanilla palette (This is the original palette of the visualizer)
+        m_palette["vanilla"].background = irr::video::SColorf(0.0,0.4,0.4,1.0);
+        m_palette["vanilla"].gridColor = irr::video::SColor(100,0,0,255);
+        m_palette["vanilla"].xAxis = irr::video::SColor(alphaLev,255,0,0);
+        m_palette["vanilla"].yAxis = irr::video::SColor(alphaLev,0,255,0);
+        m_palette["vanilla"].zAxis = irr::video::SColor(alphaLev,0,0,255);
+
+        m_palette["meshcat"].background = irr::video::SColorf(0.42,0.63,0.85,1.0);
+        m_palette["meshcat"].gridColor =  irr::video::SColor(128,128,128,100);
+        m_palette["meshcat"].xAxis = irr::video::SColor(alphaLev,234, 67, 53);
+        m_palette["meshcat"].yAxis = irr::video::SColor(alphaLev,52, 168, 83);
+        m_palette["meshcat"].zAxis = irr::video::SColor(alphaLev,66,133,244);
+    }
 #else
     DummyCamera m_camera;
     DummyEnvironment m_environment;
@@ -204,6 +232,9 @@ bool Visualizer::init(const VisualizerOptions &visualizerOptions)
         return false;
     }
 
+    // initialize the color palette
+    pimpl->initializePalette();
+
     irr::SIrrlichtCreationParameters irrDevParams;
 
     irrDevParams.DriverType = irr::video::EDT_OPENGL;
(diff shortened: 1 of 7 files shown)


src/visualization/src/Visualizer.cpp
```diff
@@ -552,4 +584,49 @@ bool Visualizer::isWindowActive() const
...
+    }
+
+    this->enviroment().setBackgroundColor(irrlicht2idyntree(colors->second.background));
+    this->enviroment().setFloorGridColor(irrlicht2idyntree(colors->second.gridColor));
+
+    // delete the frame in the origin and create a new one
+    pimpl->m_environment.m_rootFrameNode->drop();
+    pimpl->m_environment.m_rootFrameNode = addFrameAxes(pimpl->m_irrSmgr,
+                                                        pimpl->m_environment.m_envNode,
+                                                        pimpl->rootFrameArrowsDimension);
```
S-Dafarra, line 623: You are not changing the color of the axes here, aren't you?

GiulioRomualdi: ops

S-Dafarra:
> * the model is really dark (is the sun correctly placed?)

The light is placed along the z-axis by default (https://github.com/robotology/idyntree/blob/6ae3d6428e/src/visualization/src/Environment.cpp#L63-L67), so the robot is lit only in the top



> 2\. the camera is not well oriented at the startup (at least for `iCubGazeboV2_5` model)

The camera is in a specified position by default regardless of the model: https://github.com/robotology/idyntree/blob/6ae3d6428e/src/visualization/src/IrrlichtUtils.h#L387-L388 In case of ``iCubGazeboV2_5``, the initial base pose is set to the identity, that results in being reverted.

src/visualization/src/Visualizer.cpp
```diff
@@ -552,4 +584,52 @@ bool Visualizer::isWindowActive() const
...
+
+        reportError("Visualizer","setColorPalette", error.c_str());
+        return false;
+    }
+
+    this->enviroment().setBackgroundColor(irrlicht2idyntree(colors->second.background));
+    this->enviroment().setFloorGridColor(irrlicht2idyntree(colors->second.gridColor));
+
+    // delete the frame in the origin and create a new one
+    pimpl->m_environment.m_rootFrameNode->drop();
```
S-Dafarra, line 620:
```suggestion
    pimpl->m_environment.m_rootFrameNode->remove();
    pimpl->m_environment.m_rootFrameNode->drop();
```
The ``drop`` method corresponds to the destructor of a shared pointer. If the same objected is ``grab``bed somewhere else, it will not be removed. Instead, if the object calling ``drop`` is the last one having a pointer to that object, that object gets deallocated. In other words, ``drop`` just reduce by 1 the reference count. To make sure that the root frame node is removed from the scene, you have to call ``remove`` too.

S-Dafarra (approved):

src/visualization/src/Visualizer.cpp
GiulioRomualdi: In ed7ea33 I commit your suggestion however If I run `idyntree-model-view` I got a segmentation fault at
```cpp
pimpl->m_environment.m_rootFrameNode->drop(); 
```

S-Dafarra: My bad. I did not notice that ``remove`` calls ``drop``. Maybe, removing that line (the one with ``drop``) should be safe.

traversaro:
> The only two problems that we have right now are:
> 
>     1. the model is really dark (is the sun correctly placed?)
> 
>     2. the camera is not well oriented at the startup (at least for `iCubGazeboV2_5` mode

Both this problems are quite independent from the PR, right? In the sense that the PR can be merged without interfering at all with this two points, right?

traversaro (approved):

GiulioRomualdi:
What would nice to have are:
1. draw the floor as I did here. In this case, I added a cube as urdf mode that represented the floor. I don't know if it can be done in another way
       
2. If we want to replicate meshcat we should add the gradient in the background (In my opinion it is not important) let me know you think. 

Finally, I also add the default color vector to the palette

GiulioRomualdi:
@traversaro

> Both this problems are quite independent from the PR, right? In the sense that the PR can be merged without interfering at all with this two points, right?

Right

traversaro:
> draw the floor as I did here. In this case, I added a cube as urdf mode that represented the floor. I don't know if it can be done in another way

I guess there are some irrlicht methods to genereate the plane mesh, something like createMeshPlane (got it from https://stackoverflow.com/questions/40939785/irrlicht-creating-3d-plane-cube-mesh ).

GiulioRomualdi:
> I guess there are some irrlicht methods to genereate the plane mesh, something like createMeshPlane (got it from https://stackoverflow.com/questions/40939785/irrlicht-creating-3d-plane-cube-mesh ).

I would handle this in a separate PR

S-Dafarra:
It still looks good to me!

PR #874, opened 2021-06-07
Fix the visualization of boxes.

S-Dafarra:
This fixes a bug in the visualization of boxes introduced by https://github.com/robotology/idyntree/pull/845

The reason why that change was added in the first place was because I tried to add a box in the visualizer. In particular, I used a code like

```
    iDynTree::Box box;

    box.setX(0.1);
    box.setY(0.2);
    box.setZ(1.0);
    addGeometryToSceneManager(&box, nullptr, pimpl->m_irrSmgr);
```

I was expecting a box much bigger on the Z axis, but it was not the case. Indeed, it turned out that the orientation of the box was variable. Finally, by checking with ``valgrind``, I realized that ``link_H_geometry`` was read when adding the mesh to the visualizer, but it was never set. Adding the initialization of ``link_H_geometry`` solved the problem for ``valgrind`` and in the visualization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - The example file `examples/python/KinDynComputationsTutorial.py` has been fixed to work with Python 3 and iDynTree >= 2.0 (https://github.com/robotology/idyntree/pull/873).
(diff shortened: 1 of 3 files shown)


S-Dafarra:
This problem was originally discovered by @lrapetti after using the Human URDF

traversaro (approved):

lrapetti:
Thanks for the fix.

Here it is the output of `idyntree-model-view` with the current `master`
master

Here it is after @S-Dafarra changes

sdafarra

lrapetti (approved):

PR #903, opened 2021-07-21
Feature/viz viewports

S-Dafarra:
- Fixed the conversion from irrlicht pixels to iDynTree pixels.
- Fixed a typo in the environment method of the visualizer.
- Fixed the running of two visualizer instances in the same process
- Added the possibility to draw in different portions of the visualizer window and textures at the same time.
- Added the possibility to draw textures on file.
- Allow disabling the drawing on textures.
- Fixed the visualization of labels when they should be hidden.


traversaro (approved):

S-Dafarra:
Related PR to consider the change in the pixel format https://github.com/robotology/idyntree-yarp-tools/pull/16

S-Dafarra:
From the code point of view, I think this is ready to be merged

PR #911, opened 2021-08-07
[DO NOT MERGE] Test use of imported CMake targets in FindMatlab

traversaro:


traversaro:
See https://gitlab.kitware.com/cmake/cmake/-/issues/21912#note_997025 .

PR #1017, opened 2022-08-26
Avoid to compute the pseudoinverse to evaluate the external wrenches

GiulioRomualdi:
According to https://eigen.tuxfamily.org/dox/group__LeastSquares.html this should be more efficient

cc @traversaro @S-Dafarra

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- estimateExternalWrenches: Avoid to compute the pseudoinverse to evaluate the external wrenches (https://github.com/robotology/idyntree/pull/1017).
+
 ## [6.1.0] - 2022-08-18
 
 ### Added
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@
 
 cmake_minimum_required(VERSION 3.16)
 
-project(iDynTree VERSION 6.1.0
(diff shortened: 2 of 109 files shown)


traversaro (approved):

traversaro:
I just realized that in most cases we can also avoid the Qr decomposition, as in the submodels has only one full wrench as unknown, in the equation Ax = b the A is just an adjoint matrix for which the inverse is trivial.

GiulioRomualdi:
I would like to present you some results. I used `perf` to profile the new implementation with respect to the old one (the one with `pinv`)
The command I used is `sudo perf record --call-graph dwarf -p <PID> -- sleep 3` then I analyzed the data with [`hotspot`](https://github.com/KDAB/hotspot)

The following figures shows the result in case of `pinv` (the old method) and the `qr` decomposition (what the PR is proposing)
|  pinv    | QR decomposition   | 
|:---:|:---:|
| Screenshot from 2022-08-29 11-37-48 | Screenshot from 2022-08-29 11-37-37 |

Accordingly to the results the approach I proposed is 2 times faster than the original one.

traversaro:
Great, let's merge then. Given that the base branch of the PR for some reason is quite old, can I rebase on top of the latest master to run the CI?

GiulioRomualdi:
> Great, let's merge then. Given that the base branch of the PR for some reason is quite old, can I rebase on top of the latest master to run the CI?

sure!

GiulioRomualdi:
cc @gabrielenava

DanielePucci:
@gabrielenava we might consider to re-install the iDyntree for the iRonCub

PR #1051, opened 2023-02-20
Add simple tests for ModelLoader in Python

traversaro:
This should permit to debug easily your problem @S-Dafarra . At least locally it is working fine for me.


S-Dafarra (approved):
The test works on my setup

PR #1190, opened 2024-06-17
Force installing NumPy 2.0 in CI/CD when building PyPI wheels and drop Python 3.7 and 3.8 in pip packages

diegoferigo:
- Enforces the installation of NumPy 2.0 when building PyPI wheels. Otherwise, downstream projects importing both `numpy >= 2.0.0` and `idyntree` built against NumPy 1.X would fail.
- Removes support for building wheels on Python 3.7 and 3.8. This is necessary since NumPy >= 2.0.0 is not packaged in PyPI for these Python versions.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -95,7 +95,7 @@ jobs:
         run: python -m cibuildwheel --output-dir wheelhouse
         env:
           CIBW_BUILD_VERBOSITY: 1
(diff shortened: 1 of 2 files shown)


diegoferigo:
I also added a new message in CMake to make sure that the new NumPy 2.0 is found. The output produced by cibuildwheel now prints:

```
-- Linking SWIG Python bindings against NumPy '2.0.0'
```

@traversaro ready for review.

traversaro (approved):

PR #1203, opened 2024-09-17
Added possibility to change only one specific visual

S-Dafarra:
This PR adds the possibility of changing only one specific visual depending on the name.

To be tested

### DIFF
--- a/src/visualization/include/iDynTree/Visualizer.h
+++ b/src/visualization/include/iDynTree/Visualizer.h
@@ -751,6 +751,14 @@ public:
     virtual bool setLinkColor(const LinkIndex& linkIndex, const ColorViz& linkColor) = 0;
 
     /**
+     * Set the color of a single visual of a given link.
+     *
+     * This will overwrite the material of the visual, but it can be
+     * reset by resetLinkColor.
+     */
+    virtual bool setVisualColor(const LinkIndex& linkIndex, const std::string& visualName, const ColorViz& visualColor) = 0;
+
+    /**
      * Set the transparency of a given link of the model.
      *
      * This will overwrite the material of the link, but it can be
--- a/src/visualization/src/ModelVisualization.h
+++ b/src/visualization/src/ModelVisualization.h
@@ -21,7 +21,13 @@ private:
     ModelVisualization(const ModelVisualization& other);
     ModelVisualization& operator=(const ModelVisualization& other);
 
-    bool changeVisualsColor(const LinkIndex& linkIndex, const ColorViz& color, const std::string& name = "");
+    /**
+     * @param[in] linkIndex Index of the link
+     * @param[in] color The color to assign to the specified visual
+     * @param[in] visualName the name of the visual for which to change the color, or "" to set all the visual of the links
+     * @return true if all went ok, false otherwise
+     */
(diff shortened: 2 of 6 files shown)


src/visualization/src/ModelVisualization.h
```diff
@@ -20,6 +20,8 @@ class ModelVisualization: public IModelVisualization
     // Disable copy for now
     ModelVisualization(const ModelVisualization& other);
     ModelVisualization& operator=(const ModelVisualization& other);
+
+    bool changeVisualsColor(const LinkIndex& linkIndex, const ColorViz& color, const std::string& name = "");
```
traversaro, line 24:
```suggestion
    /**
     * @param[in] linkIndex Index of the link
     * @param[in] color The color to assign to the specified visual
     * @param[in] visualName the name of the visual for which to change the color, or "" to set all the visual of the links
     * @return true if all went ok, false otherwise
     */
    bool changeVisualsColor(const LinkIndex& linkIndex, const ColorViz& color, const std::string& visualName = "");
```

src/visualization/include/iDynTree/Visualizer.h
```diff
@@ -750,6 +750,14 @@ class IModelVisualization
      */
     virtual bool setLinkColor(const LinkIndex& linkIndex, const ColorViz& linkColor) = 0;
 
+    /**
+     * Set the color of a single visual of a given link.
+     *
+     * This will overwrite the material of the visual, but it can be
+     * reset by resetLinkColor.
+     */
+    virtual bool setVisualColor(const LinkIndex& linkIndex, const std::string& visualName, const ColorViz& visualColor) = 0;
```
traversaro, line 759: Not problematic, just an observation: for coherence between link and visual whould be cool to have also `setVisualColor(const std::string& linkName, const std::string& visualName, const ColorViz& visualColor)`, but this for sure do not block the PR.

S-Dafarra: Actually, there is no ``SetLinkColor`` that takes the name of the link as argument. That's why I preferred to keep using the ``linkIndex``.

S-Dafarra: Yeah, the index of the visual is a bit opaque, as it depends on the parsing of the model. So, for the scope of the PR, I preferred to keep only this version.

traversaro: Ack, but also the link index depends on the parsing, no?

davidegorbani:
I tried to modify the color of different muscles of different colors and it worked; thank you very much @S-Dafarra! Here is a picture of the human model:

Screenshot from 2024-09-17 17-01-30

davidegorbani (approved):

src/visualization/include/iDynTree/Visualizer.h
S-Dafarra: Yes. At least there is a method to get a link index given the name, but there was not a counterpart for the visuals.

davidegorbani:
> can you also try to change the color of a link whose visuals have no name?

I tried and it works fine. Here for example I set the color of the upper trunk, for which I didn't set the name of the visuals, and all the visual elements got colored.


Screenshot from 2024-09-17 17-20-25

S-Dafarra:
Great, thanks! Setting the PR ready then!

PR #1206, opened 2024-09-18
Update Python CI/CD workflow

diegoferigo:
- Fixed the folder in which artifacts are downloaded. In [#1205](https://github.com/robotology/idyntree/pull/1205) there was an extra subfolder.
- Updated the workflow to use the new PyPI Trusted Publishers.


traversaro (approved):
Thanks!

PR #1225, opened 2025-01-09
Update meshcat_visualizer.py by adding set_property

LudovicaDanovaro:
We changed this script by adding the function to set the "color" property in order to modify the color of a shape (such as a sphere) during the simulation. However it is impossible to change it in the animation


GiulioRomualdi (approved):
We tested the PR together. We were not able to make it working for the animation. However, this is something we can investigate later

PR #1247, opened 2025-06-10
Add typemaps for ContactWrench attributes in Python bindings

GiulioRomualdi:
Introduce typemaps for ContactWrench attributes,


GiulioRomualdi (commented):
This should fix https://github.com/robotology/idyntree/issues/1246

PR #1255, opened 2025-08-28
Add option for generating random models with RevoluteSO2 and fix code and tests that assumed 0-dof or 1-dof joints

Nicogene:
The RevoluteJointSO2 has been added in:
- #1254 

Fixed the `getRandomJointPositions` that was using DOFOffset instead of CoordPosOffset

This PR adds an option for generating models that includes that joints (false by default)

### DIFF
--- a/src/high-level/tests/KinDynComputationsMatrixViewAndSpanUnitTest.cpp
+++ b/src/high-level/tests/KinDynComputationsMatrixViewAndSpanUnitTest.cpp
@@ -67,7 +67,6 @@ void setRandomState(iDynTree::KinDynComputations & dynComp)
     Vector3 gravityVec;
     getRandomVector(gravityVec);
     gravity = toEigen(gravityVec);
-    gravity(2) = 0.0;  // Keep the original constraint
 
     // Use model-aware joint position generation for proper handling of all joint types
     VectorDynSize qj_idyn(posCoords);
--- a/src/high-level/tests/KinDynComputationsMatrixViewAndSpanUnitTest.cpp
+++ b/src/high-level/tests/KinDynComputationsMatrixViewAndSpanUnitTest.cpp
@@ -18,6 +18,8 @@
 #include <iDynTree/JointState.h>
 #include <iDynTree/FreeFloatingState.h>
 
+#include <iDynTree/ModelTestUtils.h>
+
 #include <iDynTree/ModelLoader.h>
 
 
@@ -47,37 +49,39 @@ int real_random_int(int initialValue, int finalValue)
 void setRandomState(iDynTree::KinDynComputations & dynComp)
 {
     size_t dofs = dynComp.getNrOfDegreesOfFreedom();
+    size_t posCoords = dynComp.model().getNrOfPosCoords();
     Transform    worldTbase;
     Eigen::Vector6d baseVel;
     Eigen::Vector3d gravity;
 
-    Eigen::VectorXd qj(dofs), dqj(dofs), ddqj(dofs);
+    Eigen::VectorXd qj(posCoords), dqj(dofs), ddqj(dofs);
 
-    worldTbase = iDynTree::Transform(Rotation::RPY(random_double(),random_double(),random_double()),
-                                     Position(random_double(),random_double(),random_double()));
+    // Use utility functions for generating random state
+    worldTbase = getRandomTransform();
 
+    // Generate random base velocity using iDynTree utility and convert to Eigen
+    Twist baseTwist = getRandomTwist();
+    baseVel = toEigen(baseTwist);
+
+    // Generate random gravity vector
+    Vector3 gravityVec;
+    getRandomVector(gravityVec);
+    gravity = toEigen(gravityVec);
+    gravity(2) = 0.0;  // Keep the original constraint
+
+    // Use model-aware joint position generation for proper handling of all joint types
+    VectorDynSize qj_idyn(posCoords);
+    getRandomJointPositions(qj_idyn, dynComp.model());
+    qj = toEigen(qj_idyn);
+
+    // Generate random joint velocities and accelerations
+    VectorDynSize dqj_idyn(dofs), ddqj_idyn(dofs);
+    getRandomVector(dqj_idyn);
+    getRandomVector(ddqj_idyn);
+    dqj = toEigen(dqj_idyn);
+    ddqj = toEigen(ddqj_idyn);
 
     Eigen::Matrix4d transform = toEigen(worldTbase.asHomogeneousTransform());
-
-    for(int i=0; i < 3; i++)
-    {
-        gravity(i) = random_double();
-    }
-
-    gravity(2) = 0.0;
-
-    for(int i=0; i < 6; i++)
-    {
-        baseVel(i) = real_random_double();
-    }
-
-    for(size_t dof=0; dof < dofs; dof++)
-
-    {
-        qj(dof) = random_double();
-        dqj(dof) = random_double();
-        ddqj(dof) = random_double();
-    }
 
     bool ok = dynComp.setRobotState(transform,
                                     make_span(qj.data(), qj.size()),
@@ -88,7 +92,7 @@ void setRandomState(iDynTree::KinDynComputations & dynComp)
 
     ASSERT_IS_TRUE(ok);
 
-    Eigen::VectorXd qj_read(dofs), dqj_read(dofs);
+    Eigen::VectorXd qj_read(posCoords), dqj_read(dofs);
     Eigen::Vector3d gravity_read;
     Eigen::Matrix4d transform_read;
     Eigen::Vector6d baseVel_read;
--- a/src/model/include/iDynTree/ModelTestUtils.h
+++ b/src/model/include/iDynTree/ModelTestUtils.h
@@ -19,10 +19,6 @@
 #include <cmath>
 #include "IJoint.h"
 
-#ifndef M_PI
-#define M_PI 3.14159265358979323846
-#endif
-
 namespace iDynTree
 {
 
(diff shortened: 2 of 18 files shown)


src/model/include/iDynTree/ModelTestUtils.h
```diff
@@ -8,15 +8,21 @@
...
 
 #include <iDynTree/TestUtils.h>
 
 #include <cassert>
+#include <cmath>
 #include "IJoint.h"
 
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
+#endif
```
traversaro, line 24: This is unnecessary, on *nix platforms `M_PI` is defined by default, while on Windows platform, `_USE_MATH_DEFINES` should be defined to ensure that `M_PI` is defined.

src/model/include/iDynTree/ModelTestUtils.h
```diff
@@ -183,14 +200,25 @@ inline void getRandomJointPositions(VectorDynSize& vec, const Model& model)
...
+            }
+            else
             {
-                vec(jntPtr->getDOFsOffset()+i) = getRandomDouble();
+                for(int i=0; i < jntPtr->getNrOfPosCoords(); i++)
+                {
+                    vec(jntPtr->getPosCoordsOffset()+i) = getRandomDouble();
+                }
             }
         }
```
traversaro, line 223: I would just use the normalizeJointPos method here, to avoid have special case (also as it is possible that in the future we will have a 2 pos coords, 1 dof joint that is not a RevoluteSO2Joint  .

src/model/include/iDynTree/ModelTestUtils.h
```diff
@@ -211,7 +239,10 @@ inline bool getRandomInverseDynamicsInputs(FreeFloatingPos& pos,
     vel.baseVel() =  getRandomTwist();
     acc.baseAcc() =  getRandomTwist();
 
-
+    // Use model-aware joint position generation instead of naive random vector
+    // This is needed for proper handling of RevoluteSO2Joint and other special joints
+    // However, we don't have access to the model here, so we can't use getRandomJointPositions
+    // This function is not used in the failing test anyway, but should be addressed separately
```
traversaro, line 245: Probably we should modify `getRandomInverseDynamicsInputs` to pass also the `iDynTree::Model` ? If we do not want to solve this here, we should open an issue, as it is really diffiult to avoid loosing track of a problem reported just with just a comment in the code.

src/model/src/FreeFloatingState.cpp
```diff
@@ -100,7 +100,7 @@ const JointDOFsDoubleArray& FreeFloatingVel::jointVel() const
 void FreeFloatingVel::resize(const Model& model)
 {
     this->m_baseVel.zero();
-    this->m_jointVel.resize(model.getNrOfDOFs());
+    this->m_jointVel.resize(model.getNrOfPosCoords());
```
traversaro, line 103: This is wrong. The `m_jointVel` and `m_jointAcc` vectors should have a size `model.getNrOfDOFs()`. For example, for a model with just a RevoluteSO2 joint, the size of `m_jointVel` should be 1, not 2.

src/model/src/Model.cpp
```diff
@@ -567,7 +567,7 @@ JointIndex Model::insertLinkToExistingJointAndAddJointForDisplacedLink(const std
 
 size_t Model::getNrOfPosCoords() const
 {
-    return nrOfDOFs;
+    return nrOfPosCoords;
```
traversaro, line 570: Good catch!

src/model/src/ModelTransformers.cpp
```diff
@@ -639,6 +640,23 @@ bool createReducedModelAndChangeLinkFrames(const Model& fullModel,
...
+            Axis rotationAxis_wrt_newLink2 = newLink2_X_oldLink2 * oldJointRevoluteSO2->getAxis(oldLink2);
+
+            RevoluteSO2Joint* newJointRevoluteSO2 = new RevoluteSO2Joint(*oldJointRevoluteSO2);
+
+            newJointRevoluteSO2->setAttachedLinks(newLink1, newLink2);
+            newJointRevoluteSO2->setRestTransform(newLink1_X_newLink2);
+            newJointRevoluteSO2->setAxis(rotationAxis_wrt_newLink2, newLink2);
+
+            newJoint = (IJointPtr)newJointRevoluteSO2;
+        }
```
traversaro, line 659: Good catch!

src/model/include/iDynTree/ModelTestUtils.h
```diff
@@ -167,6 +235,31 @@ inline Model getRandomChain(unsigned int nrOfJoints, size_t nrOfAdditionalFrames
     return model;
 }
 
+// Backward compatibility overloads
+inline Model getRandomModel(unsigned int nrOfJoints, size_t nrOfAdditionalFrames, bool onlyRevoluteJoints, bool includeRevoluteJointsSO2 = false)
```
traversaro, line 239:
```suggestion
inline Model getRandomModel(unsigned int nrOfJoints, size_t nrOfAdditionalFrames, bool onlyRevoluteJoints)
```

for back compatibility, we just need to provide the old API, not add a new attribute to a deprecated function.

src/model/include/iDynTree/ModelTestUtils.h
```diff
@@ -120,7 +137,7 @@ inline std::string int2string(int i)
     return ss.str();
 }
 
-inline Model getRandomModel(unsigned int nrOfJoints, size_t nrOfAdditionalFrames = 10, bool onlyRevoluteJoints=false)
+inline Model getRandomModel(unsigned int nrOfJoints, size_t nrOfAdditionalFrames = 10, bool onlyRevoluteJoints=false, bool includeRevoluteJointsSO2=false)
```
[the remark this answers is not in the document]
traversaro: That was exactly what I intended!

traversaro:
> I created the functions `copyFromReducedToFullDOFs` and `copyFromFullToReducedDOFs`, because in case of copying positions we need to use `getPosCoordsOffset`, for the other quantities `getDOFsOffset`, am I right?

Yes, exactly, two different functions are necessary.

src/model/include/iDynTree/ModelTestUtils.h
```diff
@@ -8,18 +8,34 @@
...
+enum JointTypes : unsigned int
+{
+    JOINT_FIXED = 1 << 0,        // 0001 - FixedJoint
+    JOINT_REVOLUTE = 1 << 1,     // 0010 - RevoluteJoint
+    JOINT_PRISMATIC = 1 << 2,    // 0100 - PrismaticJoint
+    JOINT_REVOLUTE_SO2 = 1 << 3  // 1000 - RevoluteSO2Joint
+};
+
+// Default joint types: Fixed, Revolute, and Prismatic
+const unsigned int DEFAULT_JOINT_TYPES = JOINT_FIXED | JOINT_REVOLUTE | JOINT_PRISMATIC;
```
traversaro, line 37:
```suggestion
// Simple joint types: Fixed, Revolute, and Prismatic
const unsigned int SIMPLE_JOINT_TYPES = JOINT_FIXED | JOINT_REVOLUTE | JOINT_PRISMATIC;
```

I am not a big fan of `DEFAULT_JOINT_TYPES`, as it is not clear what `default` refers to. Perhaps we can just use `SIMPLE_JOINT_TYPES`, as I think that everyone can agree that fixed, revolute and prismatic are simple, while REVOLUTE_SO2 or Spherical are not simple.

traversaro: Can we also define `const unsigned int ALL_JOINT_TYPES` with all the joint types?

Nicogene:
Just to recap,  the introduction of the RevoluteSO2Joint makes fail the `UnitTestModel` in this point:
- https://github.com/robotology/idyntree/blob/f4533323c7caa2bed91cc4a2d321e6e222dbb7a1/src/model/tests/ModelUnitTest.cpp#L276-L277

```
-8.79838
10.322
7.6331
13.3128
11.0297
19.7568

-10.8801
9.92187
10.0556
14.1364
12.5337
18.6908

vec1-vec2 :
2.0817 ( -0.2366 )
0.400127 ( 0.0387645 )
-2.42246 ( -0.240908 )
-0.823603 ( -0.0582612 )
-1.50403 ( -0.119999 )
1.066 ( 0.0539563 )
wrong el: ( . match, X mismatch):
X
X
X
X
X
X
```

src/high-level/tests/KinDynComputationsMatrixViewAndSpanUnitTest.cpp
```diff
@@ -47,37 +49,39 @@ int real_random_int(int initialValue, int finalValue)
...
+    // Generate random base velocity using iDynTree utility and convert to Eigen
+    Twist baseTwist = getRandomTwist();
+    baseVel = toEigen(baseTwist);
 
-    Eigen::Matrix4d transform = toEigen(worldTbase.asHomogeneousTransform());
+    // Generate random gravity vector
+    Vector3 gravityVec;
+    getRandomVector(gravityVec);
+    gravity = toEigen(gravityVec);
+    gravity(2) = 0.0;  // Keep the original constraint
```
traversaro, line 70:
```suggestion
```

the old line seems just a leftover, and the `// Keep the original constraint` comment does not make any sense. Let's just remove this.

src/model/src/ModelTransformers.cpp
```diff
@@ -850,6 +868,451 @@ bool createReducedModelAndChangeLinkFrames(const Model& fullModel,
     return ok;
 }
 
+// New vector-based implementation for joints with multiple position coordinates
```
traversaro, line 871: "New" with respect to what?

src/model/src/ModelTransformers.cpp
```diff
@@ -850,6 +868,451 @@ bool createReducedModelAndChangeLinkFrames(const Model& fullModel,
...
+{
+    // We use the default traversal for deciding the base links of the reduced model
+    Traversal fullModelTraversal;
+    fullModel.computeFullTreeTraversal(fullModelTraversal);
+
+    // To compute the inertia and all the other properties of the links of
+    // the reduced model, we need to build a SubModelDecomposition of the full
+    // model, in which every SubModel corresponds to a link in the reduced model
+    SubModelDecomposition subModels;
+
```
traversaro, line 889: Can we avoid the duplication? This function completly duplicates the existing `createReducedModelAndChangeLinkFrames` function.

PR #1275, opened 2025-11-25
Add feature to set frame as FloatingBase

LoreMoretti:
This PR tackles #422.

This feature has been developed following @traversaro suggestion of https://github.com/robotology/idyntree/issues/422#issuecomment-3575152858.

### DIFF
--- a/src/high-level/src/KinDynComputations.cpp
+++ b/src/high-level/src/KinDynComputations.cpp
@@ -3226,20 +3226,95 @@ bool KinDynComputations::inverseDynamics(const Vector6& baseAcc,
     // Needed for using pimpl->m_linkVel
     this->computeFwdKinematics();
 
+    // Handle transformation when floating base is a frame vs a link
+    // The input baseAcc is the acceleration of the floating base frame, but internally
+    // we need the acceleration of the base link in body-fixed representation.
+    Vector6 baseLinkAcc;
+
+    if (this->pimpl->m_isFloatingBaseFrame)
+    {
+        // baseAcc is the acceleration of the floating base frame.
+        // We need to convert it to base link acceleration (in the same representation first),
+        // then convert to body-fixed representation.
+        Transform baseLink_T_baseFrame = this->pimpl->m_baseLinkToBaseFrame;
+        Transform baseFrame_T_baseLink = baseLink_T_baseFrame.inverse();
+
+        if (pimpl->m_frameVelRepr == INERTIAL_FIXED_REPRESENTATION)
+        {
+            // In inertial representation, the acceleration vector is the same
+            // for both base frame and base link (they are rigidly connected)
+            baseLinkAcc = baseAcc;
+        } else if (pimpl->m_frameVelRepr == BODY_FIXED_REPRESENTATION)
+        {
+            // Transform body-fixed acceleration from base frame to base link
+            SpatialAcc baseFrameAcc_body;
+            fromEigen(baseFrameAcc_body, toEigen(baseAcc));
+            SpatialAcc baseLinkAcc_body = baseLink_T_baseFrame * baseFrameAcc_body;
+            baseLinkAcc = baseLinkAcc_body.asVector();
+        } else
+        {
+            assert(pimpl->m_frameVelRepr == MIXED_REPRESENTATION);
+            // For mixed representation, we need to transform properly.
+            // The mixed frame has world orientation but frame's origin.
+            // We need to convert mixed acceleration from base frame to base link.
+            //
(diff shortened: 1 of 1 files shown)


src/high-level/src/KinDynComputations.cpp
```diff
@@ -844,8 +904,34 @@ bool KinDynComputations::setRobotState(const Transform& world_T_base,
...
+        // So: world_T_baseLink = world_T_baseFrame * baseFrame_T_baseLink
+        Transform baseFrame_T_baseLink = this->pimpl->m_baseLinkToBaseFrame.inverse();
+        world_T_baseLink = world_T_base * baseFrame_T_baseLink;
+
+        // For velocity transformation, we need to be careful about the representation
+        // The input base_velocity is expressed in the frame coordinates with the specified representation
+        // We need to convert it to base link coordinates with the same representation
+
+        // Let's use the standard twist transformation
+        baseLinkVelocity = baseFrame_T_baseLink * base_velocity;
```
traversaro: I am afraid this is wrong. Here `base_velocity` could be expressed in mixed, inertial or body, but we transform it as if it was always in body. Interestingly, the early comments "we need to be careful about the representation" explicitly mention that, but then do not do that. Perhaps it could make sense to first normalize the input velocity in body representation, and then convert from base frame to base link?

traversaro: Also, shouldn't this be `baseLink_T_baseFrame` ?

LoreMoretti: yes, I am fixing them locally, I did not make it to push it yesterday.

traversaro: Ah sorry! I should have resisted the urge to comment the draft PR. : )

src/high-level/tests/KinDynComputationsUnitTest.cpp
```diff
@@ -938,6 +938,89 @@ void testSparsityPatternAllRepresentations(std::string modelName)
...
+    Transform world_H_l_sole_from_l_sole = dynComp.getWorldTransform(baseFrameName);
+    Position com_pos_from_l_sole = dynComp.getCenterOfMassPosition();
+
+    // Test that computed quantities are consistent
+    ASSERT_EQUAL_TRANSFORM(world_H_l_sole_from_l_foot, world_H_l_sole_from_l_sole);
+    ASSERT_EQUAL_VECTOR(com_pos_from_l_foot, com_pos_from_l_sole);
+
+    // Additional tests: Check that l_foot transform is consistent
+    Transform world_H_l_foot_from_l_sole = dynComp.getWorldTransform(baseLinkName);
+    ASSERT_EQUAL_TRANSFORM(world_H_l_foot, world_H_l_foot_from_l_sole);
```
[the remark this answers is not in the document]
LoreMoretti: Sure, thanks for suggesting it!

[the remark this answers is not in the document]
LoreMoretti: In part done with [2d8e3af](https://github.com/robotology/idyntree/pull/1275/commits/2d8e3af7d0f2b1877105813d4596bc5b572ff0ca)

LoreMoretti:
Now, I'd like to add more unit tests to check whether more changes are needed!

traversaro:
> Now, I'd like to add more unit tests to check whether more changes are needed!

Can you also rebase on the latest master to solve the conflicts?

src/high-level/src/KinDynComputations.cpp
```diff
@@ -844,8 +905,54 @@ bool KinDynComputations::setRobotState(const Transform& world_T_base,
 
     this->invalidateCache();
 
-    // Save pos
-    this->pimpl->m_pos.worldBasePos() = world_T_base;
+    // Handle transformation when floating base is a frame vs a link
+    Transform world_T_baseLink;
+    Twist baseLinkVelocity;
```
traversaro: Can you explain what this variable contains in a comment before the variable is introduced or in the variable name? If the base is a link frame and not an additional frame, this is exactly equal to input velocity (that is expressed in the selected velocity representation), while if the base is an additional frame, this contains the velocity of the link to which the frame is attached, expressed with the velocity representation specified in `m_frameVelRepr` . Can you explain in a comment this, and perhaps name the variable `baseLinkVelocityInSelectedVelRepr` ?

src/high-level/src/KinDynComputations.cpp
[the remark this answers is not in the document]
LoreMoretti: Addressing it in #1282

Issue #1276, opened 2025-11-25
Suspected Bug in Inertia Matrix of Kinematic Trees involving Spherical Joints


mstoelzle:
# iDynTree Bug Report: CRBA does not compute intra-joint off-diagonal mass matrix terms for spherical joints

## Summary

The Composite Rigid Body Algorithm (CRBA) implementation in iDynTree does not correctly compute the off-diagonal terms within the mass matrix block corresponding to a single spherical joint. For a spherical joint with 3 velocity DOFs, the mass matrix should contain a full 3×3 block, but iDynTree only computes the diagonal elements of this block (setting off-diagonals to zero).

## Expected Behavior

For a spherical joint with motion subspace $S \in \mathbb{R}^{6 \times 3}$ and composite rigid body inertia $I_c \in \mathbb{R}^{6 \times 6}$, the mass matrix block should be:

$$M_{jj} = S^T I_c S$$

This is a full 3×3 matrix. For a spherical joint where $S = \begin{bmatrix} 0_{3 \times 3} \\ I_{3 \times 3} \end{bmatrix}$, this computes the lower-right 3×3 block of the composite inertia.

## Actual Behavior

iDynTree's CRBA computes only the diagonal elements:

- $M_{j,j}[0,0] = S_0^T I_c S_0$
- $M_{j,j}[1,1] = S_1^T I_c S_1$  
- $M_{j,j}[2,2] = S_2^T I_c S_2$

But NOT the off-diagonal elements:

- $M_{j,j}[0,1] = S_0^T I_c S_1$ ← **missing (set to 0)**
- $M_{j,j}[0,2] = S_0^T I_c S_2$ ← **missing (set to 0)**
- $M_{j,j}[1,2] = S_1^T I_c S_2$ ← **missing (set to 0)**

## Root Cause Analysis

Looking at the CRBA implementation in `src/model/src/Dynamics.cpp`, the algorithm processes each DOF one at a time in a loop:

```cpp
// For each DOF, compute the corresponding column of the mass matrix
for (int dof = (int)model.getNrOfDOFs()-1; dof >= 0; dof--) {
    // Get motion subspace column for this DOF
    toEigen(linkJoint->getMotionSubspaceVector(..., dof, ...));
    
    // Compute F = Ic * S_dof
    toEigen(bufs.F) = toEigen(bufs.linkCRBIs(visitedLinkIndex)) * toEigen(bufs.S);
    
    // Only compute M[dof, dof] = S_dof^T * F
[...]

PR #1277, opened 2025-11-25
Fix Wrong Inertia Matrix For Spherical Joints (#1276)

mstoelzle:
This pull request introduces improvements to the mass matrix computation for multi-DOF joints, such as spherical joints, and adds a corresponding unit test to ensure correctness. The main focus is on accurately computing the coupling terms between different degrees of freedom (DOFs) of the same joint, which is essential for realistic dynamics simulation.

It closes issue #1276.

### Mass matrix computation improvements

* Enhanced the `CompositeRigidBodyAlgorithm` in `Dynamics.cpp` to compute intra-joint off-diagonal mass matrix terms for multi-DOF joints, representing the coupling between different DOFs of the same joint. This involves calculating symmetric coupling terms for each pair of DOFs within a joint.

### Testing

* Added a new unit test `SphericalJointMassMatrix` to verify the correctness of the mass matrix computation for spherical joints.

### DIFF
--- a/src/high-level/tests/KinDynComputationsUnitTest.cpp
+++ b/src/high-level/tests/KinDynComputationsUnitTest.cpp
@@ -6,6 +6,7 @@
 
 #include <iDynTree/Position.h>
 #include <iDynTree/SpatialAcc.h>
+#include <iDynTree/SpatialInertia.h>
 #include <iDynTree/SpatialMomentum.h>
 #include <iDynTree/Transform.h>
 #include <iDynTree/Twist.h>
@@ -16,6 +17,7 @@
 #include <iDynTree/FreeFloatingState.h>
 #include <iDynTree/JointState.h>
 #include <iDynTree/KinDynComputations.h>
+#include <iDynTree/Link.h>
 #include <iDynTree/Model.h>
 #include <iDynTree/ModelTransformers.h>
 #include <iDynTree/SubModel.h>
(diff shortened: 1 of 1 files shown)


traversaro:
Thanks! If you are up to, I thought of another test that we can add to https://github.com/robotology/idyntree/blob/0d97713b96689d40c53d1ab90239771595b1b325/src/model/tests/ModelUnitTest.cpp#L468 or somewhere else, that should catch any other possible error in the mass matrix for arbitrary models: if we generate a series of random model with also spherical joints and we generate a random position and velocities for these model, we can compute the total kinematic energy of the model in two ways:
* By computing the body 6D velocity v_l for each link l (via forward kinematics), and then compute the kinematic energy by summing for each link $\sum_l \mathrm{v}_l^T I_l \mathrm{v}$
* By computing the mass matrix via crba, and computing the kinematic energy $\nu^T M(q) \nu$

This is non-blocking, but it should give us quite a safe check that does not depend on any hardcoded "expected" mass matrix for a simple case that could have bugs as well.

mstoelzle:
> Thanks! If you are up to, I thought of another test that we can add to
> 
> https://github.com/robotology/idyntree/blob/0d97713b96689d40c53d1ab90239771595b1b325/src/model/tests/ModelUnitTest.cpp#L468
> 
> or somewhere else, that should catch any other possible error in the mass matrix for arbitrary models: if we generate a series of random model with also spherical joints and we generate a random position and velocities for these model, we can compute the total kinematic energy of the model in two ways:
> * By computing the body 6D velocity v_l for each link l (via forward kinematics), and then compute the kinematic energy by summing for each link 
>     
>       ∑
>       l
>     
>     
>       
>         v
>       
>       l
>       T
>     
>     
>       I
>       l
>     
>     
>       v
> * By computing the mass matrix via crba, and computing the kinematic energy 
>     
>       ν
>       T
>     
>     M
>     (
>     q
>     )
>     ν
> 
> This is non-blocking, but it should give us quite a safe check that does not depend on any hardcoded "expected" mass matrix for a simple case that could have bugs as well.

Thanks for the nice idea! I have pushed such a test!

PR #1281, opened 2025-11-26
Cleanup logic to set PYTHONPATH for Python tests

traversaro:
Fix https://github.com/robotology/idyntree/issues/1274 .


traversaro:
fyi @mstoelzle @LoreMoretti

PR #1283, opened 2025-12-11
Add algorithm to compute Coriolis matrix such that dotM = C + C^T

LoreMoretti:
It tackles https://github.com/robotology/idyntree/issues/745.

It implements Algorithm 1 of [Numerical Methods to Compute the Coriolis Matrix and Christoffel Symbols for Rigid-Body Systems](https://arxiv.org/abs/2010.01033).


traversaro (approved):

LoreMoretti:
> That is probably due to the old Eigen version on 22.04 . @LoreMoretti do you have any idea on how to quickly fix this?

Let me think about it ~1 hour

PR #1296, opened 2026-03-27
Fix conditional for PyPI publish action

flferretti:
Similarly to https://github.com/gbionics/rod/pull/73, the conditional on the CI workflow file had the wrong repository link hardcoded, so the releases were never actually published to PyPI


traversaro (approved):

flferretti:
@traversaro I added a trusted publisher with the new `gbionics` owner at https://pypi.org/manage/project/idyntree/settings/publishing/, otherwise this workflow will fail

